PR #12, opened 2018-04-01
blutil: add expression to detect test models

coolya:
In many generators we don't want to generate when we are in node or
editor tests. For that we typically check if there is any concept from
the jetbrains.mps.lang.test langauge in the original model. In order
to have this logic in a common place there is now a expression that
checks for it.


arimer:
Looks good to me, and is definitely useful!

PR #46, opened 2018-07-31
migrate mpsutil.common

fvollmer:
I tried to migrate mpsutil.common by following https://jetbrains.github.io/MPS-extensions/Migrating/.

The model checker complains about some errors, but these were already in mbeddr core.
errors


slisson:
We shouldn't move that mess to this repository. If something from there is required it should go into a new solution with a clear responsibility. If it is only required for #47, we can remove the dependency there and reject this PR.

PR #93, opened 2018-08-06
Model Merger: Language extension + runtime

krishnanm86:
The model merger allows users to specify how models in a particular language should be merged using the

"IdentityCalculatorRegistry" node:

image

Each identitycalculatorregistry contains a list of key->value entries, where each key is a concept in the language where the models need to be merged and the value specifies the base language expression that defines how the identity is calculated for that concept. These expressions contain a "paramNode" object access which refers to an object of type of concept pointed to by the "key"

Internally, the identityCalculatorRegistry generates a bunch of extension points and java classes, which are then used by the runtime to perform the merge. The merging magic happens in ModelMerger class. This class essentially parses through the input models, figures out the nodes (children/refs/props) that have been inserted/deleted/modified and performs the required changes accordingly. The runtime also takes care of positions when comparing children.

The modelmerger pull request contains the following modules:
image

1) modelmerger.runtime(solution) -> where the merging magic happens
2) modelmerger (language) -> The language defining the IdentityCalculatorRegistry rootNode and all of the stuff required to enable the proper generation of the extenstion point and classes which can be consumed by the runtime.
3) modelmerger.testLanguage -> A test language to test the model merger. It is a simple language the contains components, sub components, ports that covers all simple possibilities of references and children
4) modelmerger.sandbox -> sample IdentityCalculatorRegistry for the test language concepts
5) modelmerger.ui -> Describes a simple action that invokes the runtime backend. 
6) modelmerger.testlanguage.sandbox -> contains two models testMdlSource and testMdlDestination. A simple use case would be to invoke the modelmerger ui action on testMdlSource and choosing testMdlDestination as the target model to be merged into the source. 

image


krishnanm86:
@coolya I have removed the ui solution, added a test solution which does the stuff I used to do with the UI (for evaluation) and also changed the language to include a conceptfunction (with inputNode) instead of a simple expression as this would be required for id calculation involving complex paths (as discussed)

joao-silveira:
Some ideas from my side:
### Model merger
Please document at least the public methods. They have several parameters and it is sometimes hard to understand what is what. Besides that I find the naming a bit confusing:
`public static void matchModelsInto(model mdlSrc, model mdlDest, Map<string, IdentityCalculator> identityCalculatorReg)`
The way I understand the code it is:
**mdlSrc** is the original model we are merging into
**mdlDest** is the original model we are merging from

I would read it completly the other way around.

You could replace also: 

```
foreach nodeToRemove in nodesToRemove { 
  mdlSrc.roots(<all>).remove(nodeToRemove);
}

```
with the single liner:

`mdlSrc.roots(<all>).removeAll(nodesToRemove); `

Comments seem also a bit off to me. In example:
Comment `// Case when a root node in the source does not exist in the destination it needs to be removed from the source`
should be at least inside the if(correspondingRoot==null).

Are you sure that:
`public static void matchInto(SNode origNode, SNode importedNode, Map<string, IdentityCalculator> identityCalculatorReg);
`

Is intended to be public? The biggest problem is that it relies on the static field `mdlSource` which is set by `matchModelsInto`. If you call from the outside you don't what is the state of `mdlSource`.

Also the method `findReferenceIn` needs to be renamed and at least documented. I couldn't yet understand what it is doing.


For the **matchInto** you should also extract the content into different methods. The current state makes it not so hard to read.

joao-silveira:
Another idea from my side:
I think the actual model merger doesn't really need to know about identity registry or how the node comparators work. I think it would be nice if we could separate these two concerns.
The merger should get essentially a list of "steps" that basically would describe how to get the model from state A to state B. These steps would be executed in a single "transaction" and they could be rolled back.

I think this solution could have some advantages, at least:

- We could build different languages, comparators on top of this merging framework without having to change the actual merger (maybe extending in some scenarios)
-We could reproduce a merge multiple times. I am thinking in a case where something fails, we could roll back the merge and show the user exactly which change caused the failure. This would give a chance for the user to fix it and run again the same merge.

krishnanm86:
@joao-silveira I have incorporated all of the feedback requested barring the ones with future work and the following :
1) Having an applicable language + registering one concept per language (Reason: This would count as a future extension and should not delay the initial release of this feature)

2) Should getProperties use some more funky logic other than simply their values? I think not, because this would involve too much complexity for the language engineer when defining identities. For all of our current requirements, we have not yet faced this requirement and we should take it as the need arises. Nevertheless, I have added a todo comment in the method.

If you feel the other feedback has been incorporated satisfactorily, could you please merge it into the repo? 

Thanks in advance.

joao-silveira:
I can't get the branch to build, please:

1. Clean up the non-used dependency `de.slisson.mps.javadoc` from modelmerger.runtime
2. There are missing dependencies in the test build script. Either fix them or reload the build script.
3. Why do we need the `de.itemis.mps.modelmerger.stubs` and why do we need the jaxb jar? shouldn't this be available in the JDK?
4. The unit test `basicTests` is also failing for me. Can you please check? Also, replacing `assert` with `assert equals` in this case makes sense for me. Would make it easier to spot the expected vs output value and the assert message is much better.
5. Why didn't you use method documentation instead of adding comments in the body? This also looks weird to me.

krishnanm86:
@joao-silveira Thanks for your quick feedback. I have incorporated all of your feedback. Sorry about the issue with the build, forgot to check-in a change. 

Regarding the test failure, I am unable to reproduce it. I am running the tests with "Execute in same thread" in MPS. Please re-try this with the current latest commit and keep me posted if the test still fails.

PR #118, opened 2018-08-21
Move GrammarCells from mbeddr to MPS-Extensions

slisson:


slisson:
PR that removes GrammarCells from mbeddr: mbeddr/mbeddr.core#1981

slisson:
Try after rebuilding the sandbox language. MPS doesn't invalidate the modules that depend on a generator.

fvollmer:
Thanks, after a rebuild the test runs through.

PR #127, opened 2018-10-04
de.slisson.mps.tables: mark GridCells as big if they are big (fixes SECA-635)

abstraktor:
# Problem
GridCells break the contract that "_cells with a different node than their parents need to be marked as big and registered as such_".  This is a responsibility of the creator, e.g. handled by the `EditorManager` if created in the editor. When creating cells in other ways, `UpdaterSession#registerAsBigCell(EditorCell)` needs to be called.

Source of this contract:
image

# Steps to reproduce

1. Create a node `myNode`
1. Create a `d.s.m.tables.Table` below a separate root node
2. Add a cell with `myNode` as sub-cell of a GridCell
3. Focus the cell
4. Run the "Add margin comment" intent
5. The editor tab doesn't redraw
6. Restart MPS / the RCP
7. The editor tab doesn't redraw

# Severity
- this has lead to incompatibility with `group.margincell_review` from mbeddr.mpsutil, leading to a broken editor tab when trying to display the annotated gridcell. Restarting the application didn't help, only undoing / resetting the model
- for margincell_review, the issue can be solved by marking the inserted subcells as big and registering them, yet leads to a non-big cell marked as big, thus still breaking the contract
- I'm unsure if other languages rely on this
- This only takes effect if the node is not a descendant of the current root node, or, in other words, lives in a different editor
- This cannot be solved by changing the lookup in the margin language to walking up the node-tree (as `GraphSelection#findCellInTree` does), because this would place the comment in `myNode`'s editor rather than the grid cell's editor

# Checklist
- [x] implement hotfix in SECA
- [x] implement fix
- [x] implement automated test
image

# Links

- The corresponding internal itemis bug: https://intern.itemis.de/jira/browse/SECA-635

### DIFF
--- a/code/tables/languages/de.slisson.mps.tables/runtime/models/de/slisson/mps/tables/runtime/cells.mps
+++ b/code/tables/languages/de.slisson.mps.tables/runtime/models/de/slisson/mps/tables/runtime/cells.mps
@@ -8407,6 +8407,63 @@
             </node>
           </node>
         </node>
+        <node concept="3clFbH" id="4pEGNEjgADZ" role="3cqZAp" />
+        <node concept="3clFbJ" id="4pEGNEjgDtS" role="3cqZAp">
+          <node concept="3clFbS" id="4pEGNEjgDtU" role="3clFbx">
+            <node concept="3SKdUt" id="4pEGNEjgMl7" role="3cqZAp">
+              <node concept="3SKdUq" id="4pEGNEjgMl9" role="3SKWNk">
+                <property role="3SKdUp" value="if we are a big cell, register as such" />
+              </node>
+            </node>
+            <node concept="3clFbF" id="4pEGNEjgIIm" role="3cqZAp">
+              <node concept="1rXfSq" id="4pEGNEjgIIl" role="3clFbG">
+                <ref role="37wK5l" to="g51k:~EditorCell_Basic.setBig(boolean):void" resolve="setBig" />
+                <node concept="3clFbT" id="4pEGNEjgJvo" role="37wK5m">
+                  <property role="3clFbU" value="true" />
+                </node>
+              </node>
+            </node>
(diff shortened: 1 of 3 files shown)


abstraktor:
> I am not able to reproduce the problem in MPS 2017.3 

Maybe I can help you next week to set up a manual example of this. I am currently unsure if also shows on 2018.

> I am not able to reproduce the problem in MPS 2017.3 or MPS 2018.2. If I understand it correctly, the GridCell is selected when you invoke the intention and the margin comment is attached to the GridCell, but it cannot be found

exactly.

> because "findCellWithId" starts to search in the child of the GridCell.

Not really in the child: In the relevant call to `findCellWithId(myNode, id)`, myNode is the child node as provided in step 3. In it's first line, `findNodeCell(myNode)` retrieves its big cell. This retrieval fails if it wasn't registered as such. So we're still not evaluating the id for the subcell yet; we're just trying to find the big cell for this node. In the reproduction steps, myNode is actually located in a separate editor-component ("separate root node" in step 2).

> The correct solution would be to set the SNode of the GridCell to the one of the table.

This would be one way to solve it, yes. Still, this would require the subcell of a GridCell (let's call it content cell) to be registered as big cell, in case it differs from the parent. This sounds error-prone if we leave it to the creator of the content cell, which the creator may easily forget. Furthermore, it requires access to the parent cell when constructing the child cell to decide if it's big or not, which doesn't sound good to me.

That's why I decided for doing the big-cell registration in the tables language, and then I didn't see a need any need for changing the node of the GridCell, it is irrelevant to this issue if the GridCell or its child is big. Yet, if there is a need to change this, I'll be happily doing that.

slisson:
There is an alternative fix on the branch "tables_gridcell_snode" (MPS 2018.2).

coolya:
As far as I can see that branch was merged. Is this obsolet now?

slisson:
Yes

PR #142, opened 2018-11-23
WIP: model merger rework

coolya:
Add documentation about the merge specification as part of the documentation.

Based on the discussions we had in the office and on the whiteboard two weeks ago and the input from @krishnanm86 I have compiled this summary. This PR is the place  were we should discuss the implementation as well as the documentation. 

Please add your comment though the github code review feature directly in the markdown document. This should make it easy to address the feedback and follow the discussion. Implementation should also happen in this branch.


coolya:
Ok i took a look at github projects and it seems promising but it won't easily work with a single pull request. 

I guess we need to rethink the branching then. How about we create branch `feature/model-merger-rework` which contains the reviewed work on the model merger. When the feature is complete we merge that one into master. 

And for individual issues we create separate branches from `feature/model-merger-rework` and merge these branches with code review on a pull request back int `feature/model-merger-rework`?

coolya:
> Works, but we still need some kind of issue tracker to track dependent/exclusive tasks and their links to branches we create.

I don‘t understand this. What do you mean by issue tracker? That‘s essentially what we want to use github projects for. Please look in the projects tab on this repository there you can find a project called „model merger“. What we will use to track issues/pull requests.

docs/extensions/model-merger.md
```diff
@@ -0,0 +1,66 @@
+# Model Merger
+
+The model merge is used to automatically merge two models. Merge is not meant in the same way as in terms VCS systems like git. It‘s main use case is incrementalisation of importer at the moment but is not limited to it.
```
joao-silveira:
```suggestion
The model merger is used to automatically merge two models. Merge is not meant in the same way as in VCS systems like git.```

Comment (joao-silveira) [inline docs/extensions/model-merger.md]:
```suggestion
The model merger is used to automatically merge two models. Merge is not meant in the same way as in VCS systems like git. At the moment its main use case is incremental imports, although it is not limited to it.
```

joao-silveira: I am not sure if I interpreted correctly the last sentence here.

krishnanm86:
> > Works, but we still need some kind of issue tracker to track dependent/exclusive tasks and their links to branches we create.
> 
> I don‘t understand this. What do you mean by issue tracker? That‘s essentially what we want to use github projects for. Please look in the projects tab on this repository there you can find a project called „model merger“. What we will use to track issues/pull requests.

Perfect. This works.

docs/extensions/model-merger.md
```diff
@@ -0,0 +1,66 @@
...
+## Model Merge Specification
```
joao-silveira: Is it intentional that above it says "Model Merger"   and here "Model Merge"?

docs/extensions/model-merger.md
```diff
@@ -0,0 +1,66 @@
...
+                         name -> left | right | manual (left, right, node) -> <function returning the new property value>
+                        prop 2 -> 
+                        ---- Children ----
+                        childA [0..1] -> left | right | drop | auto | manual (left, right, node) -> <function returning the node used for the child>
+                        child [1] -> left | right | auto | manual 
+                        child [0..n] -> 
+                                               new on the right : add | drop
+                                               only existing on the the left: keep | drop 
+                                               element:  left | right | drop | auto | manual 
+```
```
joao-silveira: Don't we have yet an example for references?If not I think we can add it do the open questions.

docs/extensions/model-merger.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Partial and Complete Specifications 
+
+In many cases we want to reuse merge specifications for instance in case a concept B extends concept A we don’t want to be able to reuse the specification for A and only specify the additional aspect of the merge of B. But it should also be possible to override the behavior of the merge specification of A in B when required without respecifying the complete merge specification for A. The same applies for interface and their implementing concepts. Hence it should be possible to define a merge specification for a interface.
```
joao-silveira:
```suggestion
In many cases we want to reuse merge specifications, for instance, in case a concept B extends concept A we want to be able to reuse the specification for A and only specify the additional aspect of the merge of B. But it should also be possible to override the behavior of the merge specification of A in B when required without respecifying the complete merge specification for A. The same applies for interfaces and their implementing concepts. Hence it should be possible to define a merge specification for an interface.
```

docs/extensions/model-merger.md
```diff
@@ -0,0 +1,66 @@
...
+
+#### Complete Specification 
+
+A merge specification is complete if it defines the merge for all properties, children or references of concept and does not reuse another merge specification.
+
+#### Partial Specification
+
+A partial merge specification is a specification that does not include all properties, children or references of a concept. I reuses other complete or partial merge specifications. For instance inherited through the configuration for the base concept or by specifications of the implementing interfaces. A partial merge specification can override parts of the inherited specifications just like a complete specification can do. A partial merge specification must at least contain all the properties, children and references directly defined on the concept it is specifying for.
```
joao-silveira:
```suggestion
A partial merge specification is a specification that does not include all properties, children or references of a concept. It reuses other complete or partial merge specifications. For instance, inherited through the configuration for the base concept or by specifications of the implementing interfaces. A partial merge specification can override parts of the inherited specifications just like a complete specification can do. A partial merge specification must at least contain all the properties, children and references directly defined on the concept or interface it is specifying for. 
```

joao-silveira: > For instance inherited through the configuration for the base concept or by specifications of the implementing interfaces
What is the difference between configuration and specification here?

docs/extensions/model-merger.md
```diff
@@ -0,0 +1,66 @@
...
+A merge specification is complete if it defines the merge for all properties, children or references of concept and does not reuse another merge specification.
+
+#### Partial Specification
+
+A partial merge specification is a specification that does not include all properties, children or references of a concept. I reuses other complete or partial merge specifications. For instance inherited through the configuration for the base concept or by specifications of the implementing interfaces. A partial merge specification can override parts of the inherited specifications just like a complete specification can do. A partial merge specification must at least contain all the properties, children and references directly defined on the concept it is specifying for. 
+
+
+## Merge Specification Composition at Runtime 
+
+The runtime needs to be able to combine the merge specifications dynamically when it’s merging models. The runtime needs to construct a *complete* merge specification for each concept it tries to merge. If construction of a *complete* merge specification fails the runtime must not start the merge.
```
joao-silveira:
```suggestion
The runtime needs to be able to combine the merge specifications dynamically when it’s merging models. The runtime needs to construct a *complete* merge specification for each concept it tries to merge. If the construction of a *complete* merge specification fails, the runtime must not start the merge.  
```

docs/extensions/model-merger.md
```diff
@@ -0,0 +1,66 @@
...
+
+Given a concept A, concept B and a interface X where B extends A and B implements X. And a *complete* specification for B then the runtime is able to merge instance of B.
+
+**Special case*: Given a concept A, concept B and a interface X where B extends A and B implements X where the concept B does not define additional properties, children or references. A *complete* merge specification for X exists and a *complete* merge specification for A exists the runtime is able to merge instance of B because the information present are enough to construct a complete merge specification for B.
+
+### Precedence 
+
+A *complete* merge specification for a concept always dominates of *partial* specifications it might inherit though extensions of concepts or implementation of interfaces.
+
+Overriding (respecifying the merge for a property, child or reference) in a *partial* or *complete* merge specification always dominates over merge specifications inherited though extensions of concepts or implementation of interfaces.
```
joao-silveira:
```suggestion
Overriding (respecifying the merge for a property, child or reference) in a *partial* or *complete* merge specification always dominates over merge specifications inherited through extensions of concepts or implementation of interfaces. 
```

joao-silveira:
> Ok i took a look at github projects and it seems promising but it won't easily work with a single pull request.
> 
> I guess we need to rethink the branching then. How about we create branch `feature/model-merger-rework` which contains the reviewed work on the model merger. When the feature is complete we merge that one into master.
> 
> And for individual issues we create separate branches from `feature/model-merger-rework` and merge these branches with code review on a pull request back int `feature/model-merger-rework`?

Sounds great to me.

docs/extensions/model-merger.md
```diff
@@ -0,0 +1,66 @@
...
+### Concrete Syntax Example
+```
+Spec
+----
+<Concept> -> Id : (node) = node.name
+                         ---- Properties ----
+                         name -> left | right | manual (left, right, node) -> <function returning the new property value>
+                        prop 2 -> 
+                        ---- Children ----
+                        childA [0..1] -> left | right | drop | auto | manual (left, right, node) -> <function returning the node used for the child>
```
joao-silveira: I think we should write down somewhere what "left | right | drop | auto " means, so:
left: changes from left dominate over changes from right (?)
right: changes from right dominate over changes from left (?)
drop: as soon as a node changes, either from left or right, it is removed (?)
auto: (?)

docs/extensions/model-merger.md
coolya: I thought "merger" specification sounds strange that's why I chose "merge"

docs/extensions/model-merger.md
coolya: I agree. I will add a paragraph about the semantics.

docs/extensions/model-merger.md
coolya: Yeah a "Example" section would make sense. I think we can add it to the paragraph about the individual operations.

docs/extensions/model-merger.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Operations 
+
+For all types of merges specification entries three operations are always available: *left*, *right* and *manual*. The semantics are largely the same no matter which part of a concept they are applied to.
+
+*left*: Changes on the left side of the merge dominate over changes on the right side. The result always contains the same state as passed as the left side,
+
+*right*: Changes on the right side of the merge dominate over changes on the left side. The result always contains the same state as passed as the right side.
+
+*manual*: The manual operation allows fine grained control about the the merge result. It allows to specify the merge result as a function. The function has three arguments: _left_, _right_ and _node_. The type of _left_ and _right_ depends on the part of the concept that is merged. For properties its the value of the property on the left and the right side of the merge. For a child role it's the child node on the left and right side of the merge. The _node_ argument always contained the complete node that is merged, for properties the node owning the property and for a child role the parent.
```
coolya: I‘m not sure if the single `node` parameter is enough here. If we only pass one node we should clearly stat in which state the node is as this point. Maybe we need to pass three nodes (_left_, _right_, _merged_) where the _left_/_right_ are the parent in the state of left or rite side of the merge. The _merged_ is the node in the current state after the merge has happened, that means it might be partially merged. One thing that came to my mind is that we should document in which order the individual parts of a merge specification are executed.

joao-silveira: > . One thing that came to my mind is that we should document in which order the individual parts of a merge specification are executed.

At first sight, I think:
1. We merge all properties. The order that we merge is to be defined, but I can think of at least two options:

- We merge per property. This means we would first i.e., merge all "name" properties from INamedConcept. The order that the different "properties" would be merged would have to be defined.
- We merge per concept. This means we iterate over the concepts to be merged and for each, we merge its properties. In this case, the merge order of the concepts is to be defined. I guess the order of the properties as well.

2. We merge children. Again more or less the same questions from above are applicable.

3. We merge references. I think it makes sense to always run this as last. 

As a more general rule, I think we can either:
1. Always run "automatic" migrations first. This means that for manual migrations, the input node you mentioned would be the merged node at the state where the automatic migrations where already performed.

2. Always run "manual" migrations first, meaning that the user always relies on the original state of the nodes to be merged (Which I actually think is a more natural approach).

joao-silveira: Something else that has been bugging me is how extensible is our solution, suppose I am using a language, which I don't have access to the sources. It ships with a default merge implementation, where you see something like:

```
MyConcept -> Id : (node) = node.name
                         ---- Children ----
                         childA [0..1] -> left
```

Now the particular semantics of my domain, require me to implement a custom merge for childA, how do I do it?

1.  I extend MyConcept and provide my own merge specification.
2. Our language provides a native way to "override" the merge for a concept.

coolya: I think the questions here is how are merge specification discovered. I think haven't really talked about this at all yet. We can choose the same approach as the interpreter that the user before the merge is invoked needs to specify which merge specifications are used. 

If we make our language a full blown language aspect I think this much more complicated and we would need a notion of overriding. Though the overriding makes the runtime much more complicated.

alexanderpann:
Closing because the implementation of the modelmerger is done.

PR #149, opened 2018-11-28
modelmerger: Introducing the extended specification language

krishnanm86:
Introducing the changes to the language specification as discussed previously. This includes the addition of concepts/editors/typesystem/constraints/behaviors related to the specification of child/property strategies for each concept. The goal is to lay the groundworks for future work involving the runtime changes and generator changes that follow.

### DIFF
--- a/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
+++ b/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
@@ -2,10 +2,10 @@
 <model ref="r:fb24ac52-5985-4947-bba9-25be6fd32c1a(de.itemis.mps.extensions.build)">
   <persistence version="9" />
   <languages>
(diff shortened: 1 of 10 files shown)


coolya:
I'm a bit confused here. There is still #142 where we discuss the spec and how the language should look like and there still some open questions there. How can we review an implementation of something where we don't exactly know what we want to build?

krishnanm86:
The basic aspects like having a customizable child/property specification are not going to change and are independent of what the outcome of the open questions are. And this could be later needed for other tasks that come up. 

We can hold the pull request until other aspects of the language are discussed and implemented.

eugenschindler:
@coolya What is the idea of this pull request? It's already open for more than 2 years and should this be integrated?

coolya:
> @coolya What is the idea of this pull request? It's already open for more than 2 years and should this be integrated?

It's been a draft on how the runtime should behave and how a surface DSL for the runtime could look like. There in no implementation to be integrated (yet).

alexanderpann:
Closing because the implementation of the modelmerger is done.

PR #153, opened 2018-12-16
build: add support for incremental generation

coolya:
- passing `mps.generator.skipUnmodifiedModels=true` to all ANT scripts
  that generate MPS models. This enables incremental build by default
- add clean task: the clean tasks removes all generated classes and
  source files. This can get used to prevent incremental build or in
  other situations where all generated classes should get removed. (e.g
  MPS updates, merging).
- add rebuild task: the rebuild task first invokes clean and the the
  normal build_languages task.
- moved all scripts into "scripts" directory: gradle assumes full
  control about all files in "build" and will removed them then running
  its clean task. While we can remove the generated ant files we can't
  remove the all scripts xml because we need it to bootstrap generation
  at least until we use the new generate models functionality of the
  mps-gradle-plugin.


build.gradle
```diff
@@ -174,23 +174,19 @@ ext.artifacts_dir = '-Dartifacts.root=' + artifactsDir
 ext.pluginVersion = "-DversionNumber=" + version
 ext.buildDate = "-DbuildDate=" + new Date().toString()
 ext.extensions_home = '-Dextensions.home=' + rootDir
+ext.skipUnmodifiedModels = "-Dmps.generator.skipUnmodifiedModels=true" //incremental build support
```
arimer: @coolya So this feature is by default always on and the only way to "deactivate" this is to run `rebuild()`, correct?

coolya: Yes the rebuild task will also trigger a clean. The clean task  deletes all generated files and MPS will then regenerate them.

arimer: Its probably easier then to making `-Dmps.generator.skipUnmodifiedModels` configurable from outside.

build.gradle
```diff
@@ -174,23 +174,19 @@ ext.artifacts_dir = '-Dartifacts.root=' + artifactsDir
...
     new File("$rootDir/build/generated/$relativePath")
 }
 
-def defaultScriptArgs = [mps_home, build_dir, artifacts_dir, ext.buildDate, ext.pluginVersion]
+def defaultScriptArgs = [mps_home, build_dir, artifacts_dir, ext.buildDate, ext.pluginVersion, ext.skipUnmodifiedModels]
 
 
 
 task build_allScripts(type: BuildLanguages, dependsOn: [resolveMps]) {
-    script scriptFile('allScripts/build.xml')
```
arimer: Just out of curiosity: Why were these input/output files and directories defined here proviously?

coolya: The idea behind it was to use the incremental build feature of gradle. With gradle 5 this will no longer work because you can no longer register none existing input files as input and the script would fail.

build.gradle
coolya: I didn't see much value in configuring it when there is no artefacts MPS shall figure out generating the artefacts. Additionally the clean task comes handy in situations where you just want to get rid of all generated code. For instance when migration. Of course a `git clean -fxd´ might to the same but this was we don't blindly remove all files not in git.

PR #161, opened 2019-01-09
shadowmodels.transfor.intentions: added intention to map all properties

arimer:
When doing any in-place transformation with the shadow modells ( input element is of same type as output) the manual mapping of properties is quite cumbersome. 
To automate this an intention is implemented. 

In the case where the output node concept matches the concept of any `TransformationParameter`  the properties of this concept are mapped automatically for the output element.


slisson:
Does it make sense if it also includes child and reference links?

arimer:
Not sure. But what would we map them to? Or are you thinking about just doing a "copy" of the children and references?

slisson:
Now I understand what it does. Forget what I said.

PR #162, opened 2019-01-09
docs: add plaintext generator documentation from readme file

coolya:
Moved the documentation from the read me file into the docs subfolder so that it shows up in our web documentation page.


eugenschindler:
Should someone have a look at this? I don't know this mkdocs.yml file, so there should be no conflicts that I would mind.

PR #172, opened 2019-02-20
build: updated to latest mps-gradle-plugin

arimer:
This PR enables the new feature of [Global Defaults](https://github.com/mbeddr/mps-gradle-plugin#providing-global-defaults) provided by the latest [mps-gradle-plugin](https://github.com/mbeddr/mps-gradle-plugin).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
         maven { url 'https://projects.itemis.de/nexus/content/repositories/mbeddr' }
     }
     dependencies {
-        classpath 'de.itemis.mps:mps-gradle-plugin:1.0+'
+        classpath 'de.itemis.mps:mps-gradle-plugin:1.2+'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -183,25 +183,22 @@ File scriptFile(String relativePath) {
 
 def defaultScriptArgs = [mps_home, build_dir, artifacts_dir, ext.buildDate, ext.pluginVersion, ext.skipUnmodifiedModels]
 
+// properties for default configuration
```
coolya: I think the property name here are self explanatory enough that we don't need such a comment that stats the obvious. I would suggest we remove it or replace with something more valuable for instance a link to the documentation of these properties.

arimer: I've added the link to the mps-gradle-plugin docu to the comment.

PR #173, opened 2019-02-20
build: upgraded to gradle 5.1

arimer:
Due to the fact that we are slowly transitioning to gradle 5.x, see: [mps-gradle-plugin](https://github.com/mbeddr/mps-gradle-plugin/pull/32) I wanted to start with the mps-extensions project.

In this PR the code was updated to be conform to the latest version of gradle 5 and the `wrapper` task was executed which updated the changed files.


build.gradle
```diff
@@ -22,8 +22,8 @@ import de.itemis.mps.gradle.*
 import java.time.LocalDateTime
 
 
-task wrapper(type: Wrapper) {
-    gradleVersion '4.4.1'
+wrapper {
+    gradleVersion '5.1'
```
coolya: Why 5.1? are there specific reasons not to update to 5.2?

arimer: Na, there aren't. Just haven't seen that 5.2 is already available.

arimer: Just FYI: I'll amend the next commit.

arimer:
Any particular reason why we should **not** merge this?

PR #177, opened 2019-04-01
WIP: actions: added go to named node action

arimer:
This contribution enables a new navigation menu (`Go To Named Node`) to allow to navigate to an named node **which is not a root node**. In addition the user can  specify the scope (`ProjectScope` or `GlobalScope)` for the action.  The overall implementation is inspired by the already existing `Go To Root Node` action in MPS.

This contribution contains:

- [x] A plugin with an [UI action](http://127.0.0.1:63320/node?ref=r%3Ac5d24603-9ab7-4d81-a21c-9fa71df6366a%28de.itemis.mps.gotoactions.plugin.plugin%29%2F1380171153649074625&project=mps-extensions) contributed to the [GoTo group](http://127.0.0.1:63320/node?ref=r%3A00000000-0000-4000-0000-011c895904a4%28jetbrains.mps.ide.actions%29%2F1204991237264&project=mps-extensions) menu
- [x] A custom finder to be able to find intances of `INamedConcept`
- [x] Some [key binding](http://127.0.0.1:63320/node?ref=r%3Ac5d24603-9ab7-4d81-a21c-9fa71df6366a%28de.itemis.mps.gotoactions.plugin.plugin%29%2F1724974655066748005&project=mps-extensions) to allow triggering of the action


arimer:
In the current implementation I am (miss-?) using a language to create a custom [finder](http://127.0.0.1:63320/node?ref=r%3A04f0a748-ca55-48d2-a504-1a05f0f09c49%28de.itemis.mps.gotoactions.findUsages%29%2F6615906293313337093&project=mps-extensions) which looks for instances of `INamedConcept`. Before merging it I would like you oppinion if this is a good way to go. Because my custom finder does not require any passed **node** inside its `find query implementation` to run the underlying [ConceptInstances](http://127.0.0.1:63320/node?ref=r%3A00000000-0000-4000-0000-011c8959028e%28jetbrains.mps.lang.structure.findUsages%29%2F1197632773078&project=mps-extensions). 

Any proposals how to express this in a cleaner way are highly welcome!

coolya:
I think before we merge this we want to decide on how to include elements into this menu. 

I think an interface isn't the right way to do it. My suggestion is that we add a new language aspect that then isn't constrained to `INamedConcept`.

markusvoelter:
I agree, an interface may be too invasive. On the other hand, a whole new aspect sounds like overengineering. How about an annotation? In any case, I think it is important to be able to control the contents.

slisson:
I have a weekend project that allows the definition of interfaces and implementations in a non-invasive way. It's like extension methods but with a polymorphic dispatch. That may be an option when it's done.

arimer:
@slisson thanks for the great feedback! I'll rework the current state.

slisson:
The files of the removed language are still there. They can be removed.

arimer:
Removed not required language files.

arimer:
Any futher ideas, what we would configure in such a custom `language aspect`? 
For now I could only think of just a specified list of elements referencing `ConceptDeclarations`. Instances of these concept would then popup in the menu.

slisson:
Closing this after 8 months. Create a new pull request, if this is still relevant.

PR #219, opened 2019-05-27
WIP:DependenciesViewer-DependenciesDataGeneration into master

heikob2:
Purpose:
- enable dependencies-data-generation on ci/teamcity and send the data to a remote server
- can be executed after build and before publish
- only on ci and when system-property "system.ddgEnabledCi" = "true" is set at the teamcity-configurations

Changes: 
- added folder "./dv-ddg-client"
- adapted "**build.gradle**":
   - added code for dependencies-data-generation at the bottom
   - dependencies-data-generation dependsOn "build_languages"
   - added dependency for "publish"-gradle-task: dependsOn buildWithDdg

The "build.gradle"-file needs special attention.


heikob2:
implementation is done. @slisson could you please review?

dv-ddg-client/ddgconfig.gradle
```diff
@@ -0,0 +1,45 @@
...
+
+//relative to iets3.opensoure/build.gradle
+//seems not be applicable on ci-server
+project.ext.projects_home = './..'
+//not used for iets3
+project.ext.user_home = '-'
+
+project.ext.plugins_location = "${mps_home}/plugins"
+//only used for run_mbeddr_core
+project.ext.application_support_plugins_location = "${user_home}/Library/Application Support/MPS2018.3"
```
slisson: There is a variable `ext.mpsPluginsDir` in the main build script.

slisson:
Looks like this doesn't add any features to the MPS-Extensions project, but just executes a dependency analysis on the artifacts. If this is true, changes to this repository are not necessary and this could be done in a less invasive way (download the artifacts from the nexus repository as all other projects do).

heikob2:
Yes it does not add any new features to mps-extension. 

For the analysis, the dependencies-generator opens the mps-extensions-project (".mps"-folder) in a headless-mps-environment and calculates the dependencies for the MPS-Project.

It not uses the generated artifacts. (from the nexus repository)

slisson:
But do we need to modify each repository that we want to analyze (and copy all these jars to it)? Or can we have a separate project that has a dependency on all the projects (git submodules?) it wants to analyze.

heikob2:
This approach sounds to be the better option. Thank you.

kbirken:
I agree with the remark that the change is quite invasive. It should at least be possible to deploy all the jars on Nexus and just do some additions to the project's gradle-build. The gradle-task can then download the jars (on demand) and do the dependency analysis. How about this approach?

heikob2:
Yes. This is also possible. 👍

PR #230, opened 2019-07-29
dropdown: fixed dropdown substitutions

joao-silveira:
The previous implementation used to set the active substitution on the last leaf, which in practice
meant that generally it was activated for the most right cell.

In this implementation we want to use the outmost editor cell from the dropdown to active the
substitution. The case we have to care for, is when in example we have no value set yet.

In that case, we have to look inside the EdiotrCell_Collection and activate the substitution for the
inner cell.


slisson:
The build script in MPS needs to be updated.

Issue #241, opened 2019-09-16
The file encoding in the TextgenText_TextGen is hard wired to UTF-8


KemalSoysal:
When using the TextgenText for file output the file encoding is always UTF-8.
This should be configurable

KemalSoysal:
The version I needed the fix for is MPS 2018.2.6
Since there is no maintenance branch for the 2018.2.6, I forked the maintenance/20182 branch from the original DSLFoundry/mps-plaintextgen. You can find the proposed chang in the fork here: https://github.com/ls-it-solutions/mps-plaintextgen/commits/maintenance/mps20182
in the commits 23b04a019d5b48a363367940911dd8a1a537c00b and eac1d74c3c079b78e50ffa946ebbbf5a489ef239

JoostvanPinxten:
This sounds like a welcome addition. Can you please implement this for the master branch as well? This way this nice feature has much higher chance to be integrated in new releases of MPS Extensions.

KemalSoysal:
done with pull request #243

JoostvanPinxten:
I just noticed that this change was added to our local installation :-) I hadn't noticed that the pull request had already been completed, and we completely forgot to thank you for your contribution! 

So here goes: thanks for your contribution @KemalSoysal!

PR #243, opened 2019-09-25
#241 add encoding property defaulting to NULL

KemalSoysal:
This is causing UTF-8 to be the default character set.
Thus no migration is needed in using projects.
Enhance the textgen component for default
and when the property is set check if charset is available in the running vm.
Provide a sample for the Windows-1252 character set


eugenschindler:
Thanks for your contribution!

PR #265, opened 2019-11-21
#241 Report encoding property for MPS 2019.1.X

cmoine:


eugenschindler:
Thanks for your contribution, @cmoine!

PR #273, opened 2019-12-19
build: removed noneffective commandLine (only one per Exec possible)

wsafonov:
Required git command is executed as a first build step on the build server.


sergej-koscejev:
Wasn’t the intent here to execute both commands?

wsafonov:
> Wasn’t the intent here to execute both commands?

You would need a separate task to do it (can't have multiple commandLine per Exec). On TC there was already a separate build step doing exactly this. Because it anyway depends on TC environment due to required github token, we should keep it there.

PR #283, opened 2020-02-13
fixed wrong tooltip display issue

wsafonov:
in case the toolip is used in the editor that renders children or descender nodes
of the same of concept and using the same editor, wrong tooltip was shown for the original
node corresponding to the tooltip of the last node rendered in the original editor. This
was due to the overwritten tooltip registration per tooltip id by rendered descender nodes.
Now only the first tooltip is registered per context preventing this unintended overwrite.


wsafonov:
This is a trivial fix and can be also applied to the previous maintenance versions.

PR #313, opened 2020-08-03
collapsible.runtime: speed-up CollapsibleCell.setCollapsibleCollapsed…

### LINKED ISSUE #312: CollapsibleCell.setCollapsibleCollapsed is slow due to many editor refresh events
I have an editor with a few hundreds of collapsible cells in collapsed mode. I have implemented an action to expand-all of these collapsed cells. The action calls CollapsibleCell.setCollapsibleCollapsed(false) several hundreds of times and is very slow. The problem seems to be that a global editor refresh is done after each call. 

After debugging, it seems that the cause is the many editor-refreshes triggered (screenshot attached). Removing several lines drastically speeds-up the setCollapsibleCollapsed()

image

danielratiu:
… by removing un-needed refresh

fixes #312

### DIFF
--- a/code/widgets/solutions/de.itemis.mps.editor.collapsible.runtime/models/de/itemis/mps/editor/collapsible/runtime.mps
+++ b/code/widgets/solutions/de.itemis.mps.editor.collapsible.runtime/models/de/itemis/mps/editor/collapsible/runtime.mps
@@ -2443,6 +2443,161 @@
             </node>
           </node>
         </node>
+        <node concept="3clFbH" id="7u6DabUE_Ux" role="3cqZAp" />
+        <node concept="3clFbJ" id="7u6DabUAuTa" role="3cqZAp">
+          <node concept="3clFbS" id="7u6DabUAuTc" role="3clFbx">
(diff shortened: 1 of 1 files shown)


slisson:
Did you check, if editors containing JComponents still behave properly? If not, we can check if there are any EditorCell_WithComponent instances in the descendants before applying this workaround.

slisson:
The failing test already answers the question.

danielratiu:
@slisson thank you for the tips. I will have a look at descendants of my cell to see if there is any EditorCell_WithComponent

slisson:
What I mean is, the .setRootCell call is a workaround required for EditorCell_WithComponent. If there we can put this into an `if` to only execute it when there are any instances of EditorCell_WithComponent.

danielratiu:
I understand - how about the call  to "editorComponent.relayout()" ?

slisson:
That's part of it.

danielratiu:
I have wrapped the expensive code inside a check for descendant cells containing JComponent (EditorCell_WithComponent). Please have a look again ...

Issue #314, opened 2020-08-23
Memory leak in the diagram DSL


danielratiu:
I have a few bigger diagrams (containing several hundreds entities). When I open the container node, everything works fine. However, by closing the editor and opening it again several times, I am running our of memory.

It seems to me that in the diagram runtime libraries there is a memory leak in the static fields.

PR #315, opened 2020-08-23
Fixing memory leaks in the Diagram DSL

danielratiu:
this pull request addresses Issue #314


slisson:
Thanks for investigating this! I reviewed it and made a few changes. Can you please check if the branch [issue314](/JetBrains/MPS-extensions/tree/issue314) still does what you expect?

danielratiu:
Thank you for having a look + your changes that make the fix much more elegant. I have tested on my big diagram (opening and closing it several times) and the memory looks good.

IMO it can be merged now.

PR #316, opened 2020-08-28
Diagrams performance improvement

danielratiu:
two commits to improve the performance of diagrams


slisson:
This is not a valid optimization. What you did is splitting one big transaction (command) into multiple small ones. This feels faster because the actions executed on EDT thread are smaller and don't blocked it so long, but of cause splitting a transaction creates other problems, as the failing test shows.

PR #320, opened 2020-09-22
Don't add BorderPainter to EditorComponent if CellLayout plugin is not installed

alex-y-wd:
Currently some additional listeners are added to `EditorComponent` even when the `CellLayout` plugin is not installed.

```
editorComponent.getUpdater().addListener(ProjectPlugin.this.myUpdaterListener); 
editorComponent.addAdditionalPainter(ProjectPlugin.this.myRootCellChangeListener); 
editorComponent.addAdditionalPainter(ProjectPlugin.this.myBorderPainter);
```

This is the proposal to disable the `BorderPainter` in a case when the `CellLayout` plugin is not used as it might cause some performance issues.


slisson:
This doesn't work if the LayoutInterceptor is installed after the creation of the EditorComponent.

alex-y-wd:
Can you provide more details please? I actually didn't change the way `LayoutInterceptor` is installed.

slisson:
There are several locations where LayoutInterceptor.installWhereRequired is called. If conditionally install the BorderPainter, you have to install/remove it later if the condition changes.

alex-y-wd:
Updated, please check.

slisson:
There are still some issues. I guess it's easier to implement it myself.

alex-y-wd:
@slisson 

Is there anything I can help with this one? It is one of the major issues for us, we would be really appreciated to get rid of it as soon as possible.

slisson:
The remaining issues are:
- The DoNothingAdditionalPainter is used as a listener for rootCell changes and should always be installed
- LayoutInterceptor.installWhereRequired can be (and is) called from locations outside the celllayout plugin. This is not handled.

Feel free to give it another try if this is an urgent issue for you.

PR #322, opened 2020-10-07
Diagram: Improve performance of diagrams with many ports ...

danielratiu:
... by avoiding 

Commit 1) unnecessary cloning ov mxGeometry
Commit 2) unnecessary conversions between mxGeometry and Bounds


slisson:
The call to mxGraphModel.setGeometry with a new geometry instance is necessary to produce an event that invalidates the UI. setX/setY/... on an existing geometry doesn't produce any events.

https://github.com/slisson/jgraphx/blob/ee3ff5159969627a56330dd71bafdbae3f22ea28/src/com/mxgraph/model/mxGraphModel.java#L947

eugenschindler:
This pull request is hanging around for 3 years already. Should we integrate or delete? Or is there still something closing a decision to go forward?

PR #331, opened 2020-11-17
Uptake MPS2020.3 EAP2

ebcopur:


szabta89 (approved):
Looks good to me. Execution of automatic migrations was not even required.

Issue #218, opened 2019-05-23
Add "native" option to CR/LF/CRLF setting


eugenschindler:
In addition to CR, LF and CRLF, add also the "native" option.
Native means: do it the same way as the XML language from JetBrains MPS.

eugenschindler:
The pull request has been merged and handled by @slisson, but not automatically closed due to some administration that was not done. Closing it now manually on request of @bartvdrmeulen.

PR #333, opened 2020-11-24
plaintextgen: Add option for native line endings #218

bartvdrmeulen:
Add an new option to the line ending setting of plaintextgen that will
generate files with the native line endings.

Signed-off-by: Bart vd. Meulen <bartvdrmeulen@gmail.com>


eugenschindler:
@coolya Is like to merge this pull request but it looks like I can only close it or comment. Am I doing something wrong?

PR #347, opened 2021-02-04
Transfer langvis plugin and docs from https://github.com/DSLFoundry/m…

eugenschindler:
As discussed with @klemensschindler and @coolya, I am transferring the langvis extension from a dedicated DSLFoundry repository to the MPS Extensions.
Note: I have built locally (works) and tested a deployment of langvis in MPS 2020.3 (works as well).


eugenschindler:
@coolya would it be possible to finish (or otherwise reject, of it's not acceptable) this PR soon? I've got issues and development work stacking up for langvis which is blocked while the transfer of code and issues is in limbo.

eugenschindler:
This thing is already hanging around for 2 months. Anyone still wants it in the MPS Extensions?

eugenschindler:
Thanks a lot, Sascha!

PR #377, opened 2021-03-31
Implement abstract methods in plaintextflow concepts

klemensschindler:
As reported by @KemalSoysal and @eugenschindler: the new versions of MPS check for concepts that do not implement abstract methods. Such errors are currently reported on the plaintextflow language.

This commit should fix this by implementing the following methods in various concepts:
* Width
* Editor_View_Width
* TextGen (this method only throws an Exception, since the concepts in this language are meant to be reduced and not used by TextGen)


eugenschindler (approved):
Looks good. Thanks!

PR #389, opened 2021-04-29
Feature/add jackson lib

dbinkele:
Adds Stub Klasses for jars in 
<groupId>com.fasterxml.jackson.core</groupId> 
and 
    <groupId>com.fasterxml.jackson.dataformat</groupId>
    <artifactId>jackson-dataformat-yaml</artifactId>

. Jars are downloaded during the gradle-build from  'https://projects.itemis.de/nexus/content/repositories/mbeddr'. 
Needs to be present in 2020.1 due to customer requirements.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -126,15 +126,10 @@ ext.publishingRepository = version.toString().endsWith("-SNAPSHOT") ? snapshotRe
 
 configurations {
     mps
-    jackson
 }
 
 dependencies {
     mps "com.jetbrains:mps:$mpsVersion"
-    jackson 'com.fasterxml.jackson.core:jackson-core:2.12.3'
-    jackson 'com.fasterxml.jackson.core:jackson-annotations:2.12.3'
-    jackson 'com.fasterxml.jackson.core:jackson-databind:2.12.3'
-    jackson 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.12.3'
 }
 
 repositories {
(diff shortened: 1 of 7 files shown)


slisson:
Looks like there are not any usages of this library. It should be added to the project that uses it.

dbinkele:
> Looks like there are not any usages of this library. It should be added to the project that uses it.

@coolya Is this the right place to put the Jackson-Library to?

build.gradle
```diff
@@ -126,10 +126,15 @@ ext.publishingRepository = version.toString().endsWith("-SNAPSHOT") ? snapshotRe
 
 configurations {
     mps
+    jackson
 }
 
 dependencies {
     mps "com.jetbrains:mps:$mpsVersion"
+    jackson 'com.fasterxml.jackson.core:jackson-core:2.12.3'
```
coolya, line 134: This needs to be `2.11.*` for MPS 2020.3 since that's the version used by MPS internally to avoid strange errors at runtime.

coolya (changes requested):
@slisson the idea was to place it in MPS-Extensions because MPS ships with jackson on the class path so that we can maintain the matching version here rather than downstream. Since we do something like that with apache.commons I thought this would be the right place to put. If not we can do it downstream. What do you think?

build.gradle
slisson: Then we should just use the jars in MPS and add a solution containing the stubs for it instead of downloading them again. @dbinkele On the "Java" tab remove the jars from the "Libraries" and on the "Common" tab add the jackson jars in the "/lib" folder of MPS. MPS will use relative paths ("$MPS_HOME$/lib/...") if you do that.

Then the MPS-extensions project is the right place for it.

dbinkele: OK ... understood. But I had such a nice download solution. It seem the Apache stuff are checked in jars. Then it would make sense to download those...?!

slisson: Yes

dbinkele: @slisson Please give me a hint how to add those jars in $MPSHOME/lib in this dialog 
image

dbinkele: @coolya Some of the Apache libs checked into the repo can be replaced by later versions which are found in the $MPS_HOME/lib folder! No major version changes only minor. Makes probably sense to use those instead of the older ones!?

slisson: You have to add a different model root first.

dbinkele: Pointing to which folder?

slisson: Macros are applied automatically. Just choose the lib folder and MPS will write $MPS_HOME$/lib to the file.

dbinkele: @slisson Please have a look once more...

slisson: The download of the jackson jars is not required anymore. Revert all changes to the build.gradle file.

slisson (approved):

Issue #394, opened 2021-05-11
[langvis] indicate root node


xmoneva:
Root nodes should be marked as such.
For example, a custom spot.

![image](https://user-images.githubusercontent.com/39014043/117817012-750b9100-b267-11eb-97b5-5f5b260ffbea.png)
(plantuml required to do so: "class ConceptName << (R, lightgrey) >>")

Issue #393, opened 2021-05-11
Make extends and implements links as UML-like arrows


xmoneva:
The arrows should look like in the picture below:
 * C extends A
 * C implements B

![image](https://user-images.githubusercontent.com/39014043/117815710-0843c700-b266-11eb-9f23-3c69dba17dea.png)
(source: https://i.stack.imgur.com/4eJ33.jpg)

PR #396, opened 2021-05-11
Improve langvis rendering 

klemensschindler:
Indicate root node with custom spot: #394 
Turn on "Hierarchy" option by default, and render extended base concept when "Hierarchy" is off: #393


eugenschindler:
Perfect, thanks @klemensschindler!
Just one tiny bookkeeping tip: if you write fix #394 and fix #393, then these will be automatically closed upon merge of the pull request (as far as I know at least). Unless that was not what you were wanting to achieve. I have closed these issues now manually. In case that was not supposed to be, feel free to reopen them.

PR #398, opened 2021-05-19
Feature/add jdk to jackson stubs

dbinkele:
@slisson This stuff also has to be in 2020.1 due to CPP.....


coolya:
Please do not merge master into that maintenance branch. Only cherry pick the relevant commit(s)!

PR #400, opened 2021-05-25
Feature/add jackson stubs20201

dbinkele:
Finally works in a RCP as plugin.

### DIFF
--- a/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
+++ b/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
@@ -21,21 +21,24 @@
       </concept>
     </language>
     <language id="3600cb0a-44dd-4a5b-9968-22924406419e" name="jetbrains.mps.build.mps.tests">
-      <concept id="4560297596904469355" name="jetbrains.mps.build.mps.tests.structure.BuildMps_TestModuleGroup" flags="ng" index="22LTRF">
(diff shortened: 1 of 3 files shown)


slisson:
Screenshot 2021-05-27 at 10 07 49

This doesn't look right. In the MPS 2020.3 version these jar are not there and they shouldn't.

dbinkele:
> Screenshot 2021-05-27 at 10 07 49
> 
> This doesn't look right. In the MPS 2020.3 version these jar are not there and they shouldn't.

With these me and @wsafonov failed to use the plugin properly. When you use the stubs in some model and you run a build imports are not found..

wsafonov:
> This doesn't look right. In the MPS 2020.3 version these jar are not there and they shouldn't.

Has to be matched to the corresponding MPS version. Looks like Daniel was addressing MPS 2020.1 here, right?

slisson:
What I mean is, that no jars are packaged into the jackson plugin on the branch for MPS 2020.3.

wsafonov:
> What I mean is, that no jars are packaged into the jackson plugin on the branch for MPS 2020.3.

But this is not merged into master, but rather into 2020.1 maintenance. So everything should be ok? Then this needs to be adjusted accordingly for each MPS version .

slisson:
The plugin is supposed to provide stubs for the jars that exist in the MPS libs folder and the class loader should load the classes from the MPS jars. Packaging the jars into the plugin is wrong in all versions.

wsafonov:
Ah, I overlooked that we are packaging here already. True, no jars should be copied from the MPS/lib into a plugin. The idea was to have a plugin with references to the jars in MPS\lib.

PR #406, opened 2021-05-26
fix #405 - translate labels and annotations to origin

### LINKED ISSUE #405: Diagram: when translating to origin a diagram, the labels associated to edges and annotations remain on old position

danielratiu:


arimer:
@danielratiu pls. have in mind that master is already on MPS 2021.1, if we need these changes in 2020.3 we need to add them to the according maintenance branch.

Issue #375, opened 2021-03-30
grammarcells: migrate to new mps editor actions


bkolb:
the current implementation of grammarcells shows in several places duplicate entries in the completion menu or entries in places where they are not allowed. 

Meanwhile the MPS editors offer new functionality to associate dedicated menus on each cell. with this, it should be possible to fix all of those problems.

alexanderpann:
Closing because it was already implemented.

PR #434, opened 2021-08-10
Grammar cells migration to the transformation menu language

slisson:
When grammar cells were implemented, transformation actions were part of the actions aspect. Later MPS introduced a new transformation menu language as part of the editor. The most relevant change is that the actions aspect doesn't have access to editor while the new transformation language does.

The goal of this PR is not just to use the transformation language in the generator, but to improve the grammar cells behavior by using information available in the editor. The first step was to introduce new high level concepts that don't just make the implementation easier, but can also be used by language developers directly.

As a replacement for left and right transformations there are now the new "before" and "after" transformations. While left/right transformations only show the transformations for the current selected leaf cell and you have to include transformations from the parent explicitly, before and after transformations always include transformations from the whole cell hierarchy. They also include the transformations from the opposite side. It's more like attaching actions to the gap before/after a cell instead of a cell itself.

Then there is a cell "grammar.sideTransformation4". The difference to the other cells with a similar name is that it supports the new transformation menu language from MPS. Instead of creating a new "Transformation menu" root, you can define the menu directly inside the editor. All the entries will be shown at exactly the position of the cell (the cell itself is not visible) with all the benefits of the before/after transformations (works on the left and right side and across the cell hierarchy).

### DIFF
--- a/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
+++ b/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
@@ -5561,9 +5561,9 @@
         </node>
       </node>
       <node concept="3LEwk6" id="5kDpuTS3hez" role="2G$12L">
-        <property role="BnDLt" value="true" />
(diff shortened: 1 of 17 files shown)


slisson:
#375

PR #435, opened 2021-08-20
Add Full Extension List

alexanderpann:
- Adds a new entry "Full  Extension List" to the menu "Extensions".
- Lists all the extensions in the form of tables sorted into categories. The first columns contain the package names, the second column contains  short descriptions.

### DIFF
--- a/docs/extensions/all.md
+++ b/docs/extensions/all.md
@@ -1,0 +1,63 @@
+# Full Extension List
+
+This is a full list of all the extensions that are contained in MPS-extensions.
+
+## Baselanguage Extensions
+Name | Description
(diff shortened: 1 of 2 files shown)


eugenschindler:
Cool idea to add this list! If I understand it correctly, this list has been created manually. I would suggest to either consider an automated prices of generating this list (e.g. from the build plugin descriptions) or to add an instruction in the checklist of the  "Adding" part of the documentation so that people who add a new extension are reminded to add their extension to this full list. My personal preference would be the first solution (Automatic generation is the list) because that prevents having to update the same information manually in several places.

PR #437, opened 2021-08-24
Bug fix for fringe case in incremental model merger

JoostvanPinxten:
This PR adds a test case that fails before applying the fix to correct the insertion behavior of the incremental model merger.

If the first node in a child list in the imported model is an unmatched node (i.e., does not appear in the original parent's child listl) then the first node is inserted at the end of the child list. The expectation is that it would be inserted at the start of the list.

This PR adds a check whether it is indeed the first unmatched node, and if there are any unmatched nodes, it simply adds the new node before the first node of the child list.

### DIFF
--- a/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
+++ b/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
@@ -13383,6 +13383,11 @@
             </node>
           </node>
         </node>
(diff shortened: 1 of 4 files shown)


dbinkele (approved):

slisson:
Can you please execute the "Reload all Modules" intention in the build scripts in de.itemis.mps.extensions.build?

dbinkele:
> Can you please execute the "Reload Modules from Disk" intention in the build scripts in "de.itemis.mps.extensions.build"?

Already done

PR #444, opened 2021-08-30
Enable antialiasing in all math editor cell models

alexanderpann:
Antialiasing was only enabled for math symbols. EditorCellModels such as
math.sqrt didn't use antialising prior to this commit. Fixes IETS3/iets3.opensource#365

### DIFF
--- a/code/math/languages/de.itemis.mps.editor.math/generator/template/main@generator.mps
+++ b/code/math/languages/de.itemis.mps.editor.math/generator/template/main@generator.mps
@@ -1241,60 +1241,6 @@
                         </node>
                       </node>
                       <node concept="3clFbS" id="9L22Epae_e" role="3clF47">
-                        <node concept="3clFbF" id="2ZXaRkXmmQb" role="3cqZAp">
-                          <node concept="2OqwBi" id="2ZXaRkXmmQc" role="3clFbG">
-                            <node concept="37vLTw" id="2ZXaRkXmqTv" role="2Oq$k0">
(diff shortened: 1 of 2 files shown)


dbinkele:
@alexanderpann @kbirken  There are a lot of places where the core-VoidType still is used. Special Constraints and checking_rules for example. Therefore os-VoidType cannot replace the core-VoidType.  The simple Solution is that core-VoidType extends os-VoidType. Hereby the editor-aspect and the constraint-part of the os-VoidType are inherited. The similar parts of core-VoidType can be deleted. But the additional places where core-VoidType is used MUST be kept. Then also a migration and deprecation is not needed.

slisson:
I would move it to EditorCell_MathBase.paintContent instead of the generator. Then it's part of the runtime and not generated into all subclasses.

kbirken:
> @alexanderpann @kbirken There are a lot of places where the core-VoidType still is used. Special Constraints and checking_rules for example. Therefore os-VoidType cannot replace the core-VoidType. The simple Solution is that core-VoidType extends os-VoidType. Hereby the editor-aspect and the constraint-part of the os-VoidType are inherited. The similar parts of core-VoidType can be deleted. But the additional places where core-VoidType is used MUST be kept. Then also a migration and deprecation is not needed.

@dbinkele This comment is probably at the wrong PR? If yes, pls delete it here to reduce confusion...

alexanderpann:
> I would move it to EditorCell_MathBase.paintContent instead of the generator. Then it's part of the runtime and not generated into all subclasses.

I have moved the changes to EditorCell_MathBase. Let me know if you want any other changes.

PR #472, opened 2021-12-01
Fixed: displaying error dialog in mps-langvis causes exception

klemensschindler:
When plantuml is not installed, the mps-langvis plugin reports this using a (modal) dialog.
This causes a "AWT events are not allowed inside write action" exception to be triggered.

The issue was originally reported on youtrack: https://youtrack.jetbrains.com/issue/MPS-34018:

I am not aware what would be the write action, because the langvis plugin should not be changing the model. 
A dialog was spawned for a tool invocation, but I am not aware of it changing the model.
This looks like a false positive on the detection of a write action.

However, the change in this pull request will anyway report this error using a "message error" statement, which is more in line with the error reporting in the rest of the plugin. Besides more consistent handling, it would prevent these exceptions being thrown.


eugenschindler:
@klemensschindler what is 34018? I was trying to find maybe an issue 340, but can't find such an issue.

klemensschindler:
I see what happened. I was referring to an issue which was reported in youtrack, but of course linking would only work in github itself. I added a link to the youtrack issue in the description.

eugenschindler:
Thanks for the explanation. Looks good to me.

PR #495, opened 2022-03-01
model client use ir compiler

coolya:
I changed the JavaScript compiler for the model-api to use the new `IR` [compiler backend](https://kotlinlang.org/docs/js-ir-compiler.html). The IR compiler can produce better code for interoperability with TypeScript and JavaScript. We are creating an implementation for the INode interface that is consumed from TypeScript. (Not using the modelix model server as a storage backend but a custom implementation.)
Or implementation is a kotlin multiplatform project targeting the JVM and JS. Since we would like to provide our implementation to teams using type script we want to export richt type information for type script tooling, hence are planning to use the IR compiler backend. Unfortunately all dependencies are required to use the IR backend as well otherwise the dependency is rejected during resolving. The `IR` backend is still beta but seems quite stable from our testing.

To be able to use the `IR` backend I upgraded to the latest kotlin multiplatform plugin `1.6.10`. To prevent problem at runtime on the JVM the jvm std-lib is use with the version `1.4.31` the same as before as the kotlin plugin defaults to the same version of std-lib as the kotlin plugin when no version number is set. 
The JS implementation uses the std-lib from 1.6.10 because the `IR` backend isn't compatible with older versions of the js-std-lib. 
I'm not aware of any consumers of the model-api that are actually using JS as a target that used in production. Modelix has the model-client implementation but afaik there are no users. 
Consumers of the model-api in a muliplatform target JS would also need to switch to the IR backend after this change. I will create a PR in modelix with the required changes after this has been merged.

### DIFF
--- a/model-api/build.gradle
+++ b/model-api/build.gradle
@@ -73,7 +73,7 @@ kotlin {
         }
         jsMain {
             dependencies {
-                implementation kotlin('stdlib-js:1.4.31')
+                implementation kotlin('stdlib-js')
             }
         }
         jsTest {
--- a/model-api/build.gradle
+++ b/model-api/build.gradle
@@ -118,7 +118,7 @@ spotless {
 }
 
 task copyJarsToMps(type: Sync) {
-    from jvmJar
+    from configurations.jvmRuntimeClasspath
     from "$buildDir/libs"
     into "$projectDir/../code/model-api/org.modelix.model.api/lib"
     rename { fileName ->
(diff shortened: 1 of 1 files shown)


slisson:
In Modelix I tried switching to Kotlin 1.6.10, but in a Kotlin JVM only project Gradle failed to resolve the multiplatform library. I had to switch back to Kotlin 1.5.30. See https://github.com/modelix/modelix/runs/5051485793?check_suite_focus=true

coolya:
> In Modelix I tried switching to Kotlin 1.6.10, but in a Kotlin JVM only project Gradle failed to resolve the multiplatform library. I had to switch back to Kotlin 1.5.30. See https://github.com/modelix/modelix/runs/5051485793?check_suite_focus=true

I think that should be fixable by adding the configuration to depend on from the project explicitly. I can give that a try.

coolya:
@slisson I think got it working on this branch in modelix: https://github.com/modelix/modelix/tree/feature/use-ir-js-compiler
I can't enable the IR compiler but the project seems to build with kotlin 1.6.10 after some changes to the build.

ftomassetti:
It is nice to see the new IR compiler backend is stable enough. I am looking forward to seeing this in action!

PR #507, opened 2022-04-06
Remove deprecated tooltip language

wsafonov:
The old deprecated `de.itemis.mps.tooltips` language was still referenced in some languages requiring the `de.itemis.mps.tooltips` plugin to be installed as transitive dependency from `de.itemis.mps.editor.diagram` and `de.itemis.mps.editor.widgets` plugins.

I have removed the deprecated language, however its old runtime is still required to render tooltips for ports in diagrams. This functionality cannot be replaced by the new tooltips runtime `jetbrains.mps.lang.editor.tooltips.runtime` due to missing TooltipProvider interface - an indirection that would allow to render tooltips for ports that don't have corresponding editor cells.

We have now two options:
- retain the old tooltip runtime only for the diagrams - in this case we should probably refactor it and leave only what's required for this specific case and also move it to the diagram plugin
- clarify if we could integrate the support for TooltipProvider interface to the standard MPS tooltips runtime


alexanderpann:
`jetbrains.mps.lang.editor.tooltips` is not fully compatible with `de.itemis.mps.tooltips`:
- not working in conditional editors
- not working in ref cells ([MPS-34050](https://youtrack.jetbrains.com/issue/MPS-34050))
- not working in EditorComponents

The main reason is that the generated hint is always applied to the  root editor.
By removing our language, tooltips would not work completely in all projects after migrating because of the mentioned issues.

slisson:
The third option would be to drop support for tooltips on ports.

slisson (approved):

wsafonov:
> By removing our language, tooltips would not work completely in all projects after migrating because of the mentioned issues.

Are we aware of such projects? The tooltips language has been moved to MPS in 2020.1 version, so I agree there is a chance for  projects still on MPS 2019 or below to run into these problems after the migration. But on the other side many projects have already switched to using the new language. So if there were blocking issues they would have come up till now.

PR #513, opened 2022-04-21
Backporting "Enrich repository concepts" to MPS 2020.3

ftomassetti:
I would need to backport this feature to MPS 2020.3


ftomassetti:
I think this should be a pretty straightforward PR, as it is just about backporting features to a previous version of MPS. It only affects the model-api files, so I think that @slisson would be the best person to review it.

I hope I can get this merged, so that I can fix some issues in Modelix (https://github.com/modelix/modelix/issues/155) for MPS 2020.3 and all successive MPS versions

PR #537, opened 2022-08-05
[breaking] Enforce @NotNull annotations at runtime

sergej-koscejev:
Switches the default compiler to IntelliJ (`<javac2>` Ant task). Javac2 instruments code to ensure that `@NotNull` variables are never null at run time.

This is a breaking change because code that used to work before may start throwing exceptions now. However, it is likely that such code didn't work before anyway and it will now throw exceptions earlier in the call stack.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ model-api/build
 code/model-api/org.modelix.model.api/lib
 scripts/.mps-caches
 .idea/modules
+
+# Generated during the build
+/kotlin-js-store/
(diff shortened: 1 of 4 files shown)


sergej-koscejev:
@slisson modelix/mpsbuild needs an option to generate javac2 tasks to remain consistent with the MPS-generated builds.

sergej-koscejev:
modelix/mpsbuild no longer used in the build so the comment about it doesn't apply

alexanderpann (approved):
Then I don't see a reason why we can't merge anymore.

PR #550, opened 2022-09-19
diagram: run autolayouter if no layout data exist and run layouter in background

### LINKED ISSUE #240: [diagram] Layout not happening for the first time
Create a state machine diagram in the implementation module.
You can see the state and the events are not layout-ed for the first time toggling.
The same thing happens with all nodes using the CellModel_Diagram concept.
Now: 
image
Before:
image

alexanderpann:
Fixes #240 The layouter now also runs in the background and can be canceled in the right bottom corner.
I've also removed the multiple calls to the layouter as this code is probably not needed anymore.
Two open questions:

- Should the layouting be really done automatically, or should there be a setting in the editor to enable/disable it?
- Do we need a configurable timeout for the layouting? In the past, the layouter froze the editor but now it might be a non-issue.

### DIFF
--- a/code/diagram/languages/de.itemis.mps.editor.diagram.demo.activity/languageModels/editor.mps
+++ b/code/diagram/languages/de.itemis.mps.editor.diagram.demo.activity/languageModels/editor.mps
@@ -272,6 +272,7 @@
       </concept>
       <concept id="6554619382999975769" name="de.itemis.mps.editor.diagram.structure.Content_GenericElementQuery_OuterNode" flags="ng" index="23r2z0" />
       <concept id="1110129820007229393" name="de.itemis.mps.editor.diagram.structure.CellModel_Diagram" flags="ng" index="27vDVx">
+        <property id="4706276119306323403" name="runAutoLayoutOnInit" index="1ju4zT" />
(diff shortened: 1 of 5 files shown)


slisson:
DiagramIdeaActionsUtil.autolayout layouts the "active" diagram. This not guaranteed to be the one without any layout data.

alexanderpann:
@slisson I've fixed the mistake. I was just told that running the layouter can take very long (> 30 seconds) for very big diagrams in some projects. That's why I decided to use the PR to also fix this issue and run the layouter in background tasks that are not blocking except when the model is written. The tasks can be canceled by the user. What do you think of the implementation? I had to synchronize some methods so that the 5 tasks are run after each other. I also place the logic for the background execution in the elk layouter directly because otherwise I could only community the task cancellation through the IntelliJ message bus, which might not be ideal. Is there any simplification possible?
(I'll fix the tests later)

alexanderpann:
There is now also an optional setting in the editor of the diagram (run autolayout on init) which can trigger the auto-layouting when no layout data is present. It is set to false by default so that the old behavior is not changed.

slisson (approved):

PR #552, opened 2022-10-04
Migrate to MPS 2021.3.1

alexanderpann:
I've included the missing changes from 2021.1.4 and 2021.2.5 and merged all three existing 2021.3 branches into this new one. I also had to disable the MPSI-36 flag in the build solution of the languages because I couldn't build the project locally while it was enabled.

What should be do with the jackson stubs that are now part of lib/3rd-party.jar?. We could drop the stub solution because creating stubs for the full jar (~140MB) doesn't seem like a good idea to me.


slisson:
We usually use the `mps/*` branches for the migration and there is already one for MPS 2021.3. I merged the `mps/2021.3` branch into `migration/mps2021.3` to not lose these changes.

slisson (approved):

slisson:
We can replace the jackson solution with a new one that includes the whole 3rd-party.jar. Because of lazy loading of models and classes only those parts that are actually used will be loaded into memory.

Issue #91, opened 2018-08-06
Diagrams: Boxes can be moved out of nested diagrams


kbirken:
From mbeddr.core created by [slisson](https://github.com/slisson) : mbeddr/mbeddr.core#1413

PR #549, opened 2022-09-12
diagram: layout edges from port of subdiagram to box in subdiagram

pavel-nedvedicky:
Set hierarchy of nodes for elk layouter, which fixes issue with edges from port of subdiagram not being layouted.

**Before:** 
image

**After:**
image

PR #554, opened 2022-10-07
diagram: fixes for diagram issues related to dragging elements outside of their parent's boundary

alexanderpann:
This PR at least partly fixes https://github.com/JetBrains/MPS-extensions/issues/91 and the following issues:
When dragging an element (or using the arrow keys) to the left/top of the diagram, the diagram normally gets extended and the move preview didn't respect the new translation of the diagram view. In addition, these elements appeared outside the diagram when pressing the reset scale and transform button because of issues with negative coordinates.
Now, this should work in addition to moving the full subdiagram instead of moving an element of the subdiagram to an arbitrary location in the diagram to keep the element within the subdiagram's boundary. When dragging the element to the left or bottom of the diagram, the width/height of the subdiagram gets automatically extended.

Unfortunately for nested diagrams, these changes do not work together with PR https://github.com/JetBrains/MPS-extensions/pull/549. For this cases, an additional change to the ElkLayouter is probably necessary.

### DIFF
--- a/code/diagram/solutions/de.itemis.mps.editor.diagram.runtime/models/de/itemis/mps/editor/diagram/runtime/model.mps
+++ b/code/diagram/solutions/de.itemis.mps.editor.diagram.runtime/models/de/itemis/mps/editor/diagram/runtime/model.mps
@@ -10802,6 +10802,161 @@
         </node>
       </node>
     </node>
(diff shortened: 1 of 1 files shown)


slisson:
The adjustment of the subdiagram boundaries still didn't work properly. I reimplemented it. Please check if it now behaves properly.

slisson (approved):

PR #560, opened 2022-11-16
[Breaking] Build script: check paths in modules.xml

alexanderpann:
This PR adds a check to the build script for the paths in the modules.xml and adds a precheck task for registering other checks that are run before every other task.
Do we need any other pre-checks at the moment?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -237,8 +237,6 @@ ext["itemis.mps.gradle.ant.defaultScriptArgs"] = defaultScriptArgs
 ext["itemis.mps.gradle.ant.defaultScriptClasspath"] = buildScriptClasspath
 ext["itemis.mps.gradle.ant.defaultJavaExecutable"] = new File(jdk_home, 'bin/java')
 
-def prechecks = tasks.register("prechecks")
-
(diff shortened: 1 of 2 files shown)


slisson:
I would replace the task dependency between "build_allScripts" and "checkModulePaths" with one between "check" and "checkModulePaths".

slisson (approved):

alexanderpann:
@slisson It makes sense to add it to checks although now such an obvious check is done as the last step. I am confused about the task hierarchy. Why is `run_test` not executed as part of `check` as in every normal gradle script? `assemble` should be called  before `check` anyway, so `build_languages` should be executed before `run_tests` right? `packageExtensions` currently depends on `run_tests` which is also strange. There is also a task `packageAllScripts` which is not executed at all:
 
```
:build                                    
+--- :assemble                              
|    `--- :packageExtensions               
|         `--- :run_tests                  
|              `--- :build_languages (default task)     
|                   `--- :build_allScripts
|                        +--- :copyModelApi 
|                        `--- :resolveMps   
`--- :check                                
     `--- :checkModulePaths  
```

PR #562, opened 2022-11-21
Support for line numbers

alexanderpann:
This is an implementation for [MPS-33595](https://youtrack.jetbrains.com/issue/MPS-33595/Number-lines-in-MPS-editor). They are disabled by default but can be enabled through preferences -> other settings -> LineNumbers.
Intersecting lines are merged by default. That's why the `root` tree node doesn't have it's own line because it overlaps with the child node on the last line. Folding and zooming are also supported. There is also a new action `Go to Editor Line` to jump to a line in the currently open editor.

Screenshot 2022-11-21 at 07 15 26
Screenshot 2022-11-21 at 07 15 46

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -237,9 +237,28 @@ ext["itemis.mps.gradle.ant.defaultScriptArgs"] = defaultScriptArgs
 ext["itemis.mps.gradle.ant.defaultScriptClasspath"] = buildScriptClasspath
 ext["itemis.mps.gradle.ant.defaultJavaExecutable"] = new File(jdk_home, 'bin/java')
 
+tasks.register("checkModulePaths") {
+    def mps_dot_folder = ".mps";
+    def modules_xml_location = "modules.xml"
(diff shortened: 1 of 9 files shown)


slisson:
In IntelliJ I can right click on the left side of the editor and choose "Show Line Number" from the context menu to enable them. I think this is the more intuitive place instead of using a preference component.

slisson (approved):

alexanderpann:
I think there is still a bug where the collapsed lines are not skipped, or the information is not correctly saved in the user object. I've managed to have this bug in LineNumberUtils with some collapsed cells. Maybe it has to do with opening a second project or updating the line numbers only for visible editor component, but I'm not quite sure.

slisson:
This happens when a cell is initially collapsed. Then there is no cached ILineList object for the expanded state.

slisson:
Since we don't have any more time for this, let's merge the current state and create issues for what is not working yet properly. The current state is already useful enough.

PR #580, opened 2023-02-07
Feature: HTML Swing-Component

alexanderpann:
Screenshot 2023-02-07 at 13 36 26

Screenshot 2023-02-07 at 13 40 44

As part of a performance investigation, I've created this experimental custom swing component that supports a small subset of HTML. With a few additional lines there could also be rudimentary support for [RTF](https://en.wikipedia.org/wiki/Rich_Text_Format) in case anybody needs that. This cell can be used, for example, for long read-only explanations instead of using many single MSP editor cells.

Unfortunately,I couldn't quite figure out the correct relayouting of the cell. I think there might still be issues where the content of the JComponent gets cuts off on the right side, that's why the PR is still in draft mode.

Not sure if we really need it, as it doesn't integrate that nicely with other editor cells and using some old HTML standard for writing messages is also not that great. At last, the approach is more lightweight than my full [JCEF integration](https://github.com/alexanderpann/mps_jcef_minimal).


slisson:
> Unfortunately,I couldn't quite figure out the correct relayouting of the cell. I think there might still be issues where the content of the JComponent gets cuts off on the right side, that's why the PR is still in draft mode.

I fixed that one for you.

slisson (approved):

PR #584, opened 2023-02-20
Don't trim intentions-menu entries that are not grouped

AlexeiQ:
This PR is a bugfix for an accidental trimming of intentions-menu entries that are not grouped and don't have a sub-menu, which was introduced in #577 

# Bug Description
For Intentions without an associated group, and no sub-menu, the first character of the intention description is cut off: 
2023-02-20_16h47_15

# Bug Fix
For intentions without a group an empty string is returned as `groupName`, which is now considered:
2023-02-20_17h11_05

# Further Changes
- Improved the sandbox language with some more examples


alexanderpann:
I've already fixed the issue in https://github.com/JetBrains/MPS-extensions/commit/f65215d935ff09fcb38b3f3c0b0ae51b90994d03. Can you check if your Pr is still needed?

AlexeiQ:
> @AlexeiQ I've already fixed the issue in [f65215d](https://github.com/JetBrains/MPS-extensions/commit/f65215d935ff09fcb38b3f3c0b0ae51b90994d03). Can you check if your PR is still needed?

@alexanderpann Sry, I started before your fix and missed it today :/ 
Nevertheless I would like to change your implementation to reduce complexity. The "+1" aka offset should be used just to cut off the `:`. In case the intention doesn't have a group name. The "cutting off" should not be considered at all. There should also be no trimming for intentions without descriptions. At least not mixed up inside that code-line, in my opinion.

alexanderpann (approved):
LGTM. BTW: We are also affected by the automatic removal of underscores in the intention description the way it works in MPS: https://jetbrains-mps.slack.com/archives/C3YUV3YK0/p1676819421949179

PR #591, opened 2023-03-14
Refactor/jbr17 cleanup

arimer:
- This PR proposes some minor cleanups to the usage of the downloadJBR taks. 
- It also removes the no longer required `jdk_home` resolution logic as well, because the path to the _downloaded java executable_ is exposed via `javaExecutable` on the task with the name `downloadJbr`.


arimer:
@alexanderpann I am not which target branch should be picked. Please advice.

alexanderpann:
I've already explained it but this system variable is needed for JetBrains for their build server.

arimer:
Maybe it is worth while adding a comment to the code, to exlain this as well.

PR #593, opened 2023-03-20
Update plugin names and descriptions

alexanderpann:
@slisson please check if you are fine with how and where you are attributed as the vendor.


slisson:
Make it consistent and remove my name.

PR #603, opened 2023-03-28
Allow overriding layout storing

HeikoBecker:
This PR adds a boolean override flag to the `saveDiagramLayout` function.
If it is set to true, the diagram layout is stored, no matter how the saving is configured  in the diagram editor. This allows properly storing layouts via user-intentions.


slisson:
Create a new method with this additional parameter instead of changing the existing one. Maybe the parameter is then not even needed and the method can be name "forceStoreLayoutData" or something like that.

HeikoBecker:
> Create a new method with this additional parameter instead of changing the existing one. Maybe the parameter is then not even needed and the method can be name "forceStoreLayoutData" or something like that.

Done. I force-pushed to get rid of the bad implementation.

slisson (approved):

PR #610, opened 2023-04-15
[Breaking] Celllayout: fit the grid when layouting the root layoutable container

alexanderpann:
With this change, the width of cells with pushX/growX is the maximum of the text width (`preferences->Editor->MPS Editor->text width`) and the widest cell in the editor. The text width was not considered on purpose in the past. That means that, for example, horizontal lines had the width of the widest cell in the editor which doesn't make any sense to me. What's the point of having a maximum width for the editor if it is not always considered in the maximum width calculation? Was there any reason to disable the fitting for the preferred inner size? I couldn't find any examples where this causes layout issues, so I changed it.

Before:
Screenshot 2023-04-15 at 20 05 54
After:
Screenshot 2023-04-15 at 20 13 24

The lines might appear a bit long in this example because I have set the text width to 180. With a more standard text width like 100 it looks normal.

Screenshot 2023-04-15 at 20 06 38

You can even restore the old behavior by placing pushX/growX: false commands at specific levels in the editor. In the following example, the content collection of the test suite has the flags set, which makes the red lines the width of the content. The black line after the header has the full editor text width:

Screenshot 2023-04-15 at 22 11 41


slisson:
I'm not sure if this behavior is always desired. At least changing the return value of `getPreferredInnerSize` is not correct, but just a hack. Somewhere in the `LayoutEngine` class would be the right place for this change, preferably configurable.

alexanderpann:
I am closing in favor of #862.

PR #626, opened 2023-05-15
MPS.ThirdParty: don't copy app.jar to avoid class loading conflicts

alexanderpann:
Let's try fixing this stub again. I've tested it by installing the plugin in MPS. It seems to work.


alexanderpann:
Alright. I've removed it. Is it fine now?

slisson:
Yes, looks good.

PR #627, opened 2023-05-16
Linter + model check + folder structure fixes (2021.1)

alexanderpann:
This branch is experimental and shouldn't be merged without thorough checking. I might have broken something.

- I've enabled 10 linters from [mps-qa](https://github.com/mbeddr/mps-qa)
- The model checker is enabled and checks all models (part of check task)
- I've fixed the folder structure for languages that the linter complained about
- I've inserted one model checking error on purpose for testing
- I've fixed some model check errors and have just suppressed the other errors so that I can get the linter working
- I've also added all test solutions to the build script so that they are executed
- Now, also all 320 tests are executed.

One disadvantage of the current approach is that you need to insert your GitHub credentials into the gradle.properties file so that you can download the linter dependency. I haven't checked how much the check task affects built time.


alexanderpann:
Small explanation why the linter "not_used_module_dependencies" was overwritten: the original linter didn't exclude dependencies with scope "design" *and* "generates into". They should not be considered unused dependencies.

slisson:
We should merge and test this change with the branches for newer MPS version first, before merging this PR.

alexanderpann:
Yes, I uploaded the artifact after I created this PR, we should probably change that.

sergej-koscejev:
Closing this because we agreed not to maintain 2021.1 any longer.

PR #648, opened 2023-07-31
Upgrade minor version to MPS 2021.3.3

HeikoBecker:


alexanderpann (changes requested):
Please also update the README. Other than that, 2021.3.3 should not have any incompatible changes, so the change should be fine.

HeikoBecker:
> Please also update the README. Other than that, 2021.3.3 should not have any incompatible changes, so the change should be fine.

Thanks for noticing. I have updated the README.

alexanderpann (approved):

PR #651, opened 2023-07-31
Remove old nexus repository and update configuration

KemalSoysal:
remove old nexus repository,
update usage of de.itemis.mps.gradle plugin according to https://github.com/mbeddr/mps-gradle-plugin/tree/v1.x#configuring-the-plugin-repository

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,6 +1,5 @@
 pluginManagement {
     repositories {
-        mavenLocal()
         maven { url 'https://artifacts.itemis.cloud/repository/maven-mps' }
 
         // Need to manually include the default Gradle plugin portal repository when overriding the defaults.
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,5 +1,6 @@
 pluginManagement {
     repositories {
+        mavenLocal()
         maven { url 'https://artifacts.itemis.cloud/repository/maven-mps' }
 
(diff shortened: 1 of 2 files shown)


settings.gradle
alexanderpann, line 1: @sergej-koscejev can you clarify the difference between the build block and the plugin management block? In the README of the mps-gradle-plugin we have for 1.x and 2.x two different descriptions. the buildscript block is enough, has the pluginManagement block any advances?

sergej-koscejev: There should be no semantic difference but using `plugins` and `pluginManagement` is more idiomatic (higher-level) than the `buildscript` block. I think we could/should remove the `buildscript` block entirely.

sergej-koscejev (approved):
I would also remove the `buildscript` block as mentioned in one of the comments, but I'm going to approve this because the changes are correct and it's a step forward.

settings.gradle
slisson: One important difference is that inside the pluginManagement block you have access to the values defined in the gradle.properties files, which can be used to load plugin versions from them.

KemalSoysal: > Don't. It makes builds unreproducible. If it's added, it should be conditional on some property and/or combined with a repository filter: https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:case-for-maven-local

pushed required change with removal of mavenLocal

PR #652, opened 2023-07-31
Fixing build

AlexeiQ:
- Backported from 21.3:  see https://github.com/JetBrains/MPS-extensions/pull/646
- Fixed the `checkModulePaths`-gradle task, so correct paths are used on windows machines
(Since the `rootDirectory` variable contains the path with windows file-separators and `replaceAll()` uses a regex, all the windows-file-separators are dismissed, leading to wrong paths for all modules in the modules.xml. The regular `replace()` does the same replacement, but without the regex, keeping the windows-file-separators in place. Because the separators are still used based on the OS, this will not influence other OS.)
- Backported  from 21.3:  see https://github.com/JetBrains/MPS-extensions/pull/651

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,10 +1,3 @@
-//will pull the groovy classes/types from nexus to the classpath
-buildscript {
-    repositories {
-        mavenLocal()
-        maven { url 'https://artifacts.itemis.cloud/repository/maven-mps/' }
-    }
-}
 plugins {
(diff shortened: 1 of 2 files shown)


build.gradle
KemalSoysal, line 1: module.xml does contain only one macro `$PROJECT_DIR$` per modulePath.
This makes the replace method being suitable for the root path replacement.

KemalSoysal (approved):

PR #656, opened 2023-08-04
Cascading merge from 2021.1 into 2021.2 with migration maintenance

AlexeiQ:
- Cascading merge from MPS 2021.1 into 2021.2 (resovling conflicts)
- Execute migrations via migration assistant (after `git clean -xdf` , `./gradlew run_tests` and `make project`)
- Executed rerunnable migrations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,6 @@
+# Changelog
+
+All notable changes to this project are documented in this file.
+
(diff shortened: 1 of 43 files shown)


KemalSoysal:
Cloned, built result is without any migration popups. Executing re-runnable migrations does not produce changes.

PR #657, opened 2023-08-10
Editor pagination dsl

amin-ait:
Children with cardinality `[0..n]` and `[1..n]` are displayed in the editor through collection cells.
Since the collection cell renders all the nodes, If we have a long list of nodes, the resulting editor might be slow.

This pull request adds a `pagination` language, that contains a `paginated` cell that can be applied to a collection cell.
As a result in the rendered editor the collection is paginated, meaning that the list of nodes is split in pages and only elements within a selected page are displayed.

This should in theory render long list of nodes more efficiently.
An example of the resulting editor is shown in the screencast below, we display a list of 1000 nodes in pages of size 100.

[Screen Recording 2023-08-10 at 14.20.35](https://github.com/JetBrains/MPS-extensions/assets/10254994/9ad1859f-996f-47da-b59d-398d6171243c)


HeikoBecker:
Some comments on the PR:

** de.itemis.mps.editor.pagination.test **

- I have tried running the tests in `de.itemis.editor.pagination.test` but they all fail to run with an exception. If they are meant to work, please fix them. Otherwise they should be deleted.

** de.itemis.mps.editor.pagination.runtime **

- Is it intended that the `hasNext` function always returns true in the `Page` object? ([node url](http://127.0.0.1:63320/node?ref=r%3A56b5798b-fb73-40ab-987e-695e19f87b03%28de.itemis.mps.editor.pagination.runtime.pages%29%2F5460816232793608499))
- same as above for the class `PageNumber`
- I am not sure whether the "hashCode-equals" contract is valid for the `Pages` object (`ob1.equals(ob2) ==> ob1.hashCode == ob2.hashCode`)
- What is the difference between these two functions in `PagesUserObject`:
image

** de.itemis.mps.editor.pagination **

- I could not find any use for this function in `Paginate_CellModel_Behavior`: 
image
- There are some warnings related to `childNode` in the generator. I could fix them by introducing an intermediate node:
image

It might be worth checking why this error appears in the first place?

dodoga93:
Thank you for the review 🙂
> **de.itemis.mps.editor.pagination.test**
> 
> * I have tried running the tests in `de.itemis.editor.pagination.test` but they all fail to run with an exception. If they are meant to work, please fix them. Otherwise they should be deleted.
> 
There was an issue when running all of them that I reported  [here](https://itemis.slack.com/archives/C64MLPHEJ/p1695217868521499).
I replaced the `BTestCase` nodes with `NodeTestCase`s and now it should be working.

> **de.itemis.mps.editor.pagination.runtime**
> 
> * Is it intended that the `hasNext` function always returns true in the `Page` object?
> * same as above for the class `PageNumber`

Yes, this is because I was trying to see if there was a common interface between the 3 classes `PageNumber`, `Page` and `Pages`.
I've made now this interface explicit by adding the interface [`PageIterator`](http://127.0.0.1:63320/node?ref=r%3A56b5798b-fb73-40ab-987e-695e19f87b03%28de.itemis.mps.editor.pagination.runtime.pages%29%2F88315915626305945).

> * I am not sure whether the "hashCode-equals" contract is valid for the `Pages` object (`ob1.equals(ob2) ==> ob1.hashCode == ob2.hashCode`)

I've improved the code for `Pages.equals` by using the same function `getCurrentElementsCount` as in `hashcode`, with the current implementation there could be collisions.

> * What is the difference between these two functions in `PagesUserObject`:
>   image
> 
The method `Pages.size` returns the number of pages needed for the collection, the method `getPageSize` returns the size of a single page, how many elements a page can contain.

> **de.itemis.mps.editor.pagination**
> 
> * I could not find any use for this function in `Paginate_CellModel_Behavior`:
>   image

I removed it.

> * There are some warnings related to `childNode` in the generator. I could fix them by introducing an intermediate node:
>   image
> 
> It might be worth checking why this error appears in the first place?

The error is because, concepts like `childNode` use the context to calculate their type ([`typeof_ConceptFunctionParameter_childNode`](http://127.0.0.1:63320/node?ref=r%3A00000000-0000-4000-0000-011c8959029a%28jetbrains.mps.lang.editor.typesystem%29%2F1182235413427)), in particular it will use the type of the list of nodes that the collection cell shows, this in the generator doesn't have a type, so the type for `childNode` remains unsolved as well.
I've added an intermediate node as you suggested.

HeikoBecker (changes requested):
I have fixed the build scripts and included the tests in them.

Two more comments before we merge:

Can you add a comment to the PaginationListener on which other modellistener its implementation is inspired by for future reference?
And finally, I think we should remove this notification:
image

amin-ait:
Thanks for fixing the build script, I forgot to do that.

> Can you add a comment to the PaginationListener on which other modellistener its implementation is inspired by for future reference?
> And finally, I think we should remove this notification:

I did that in the last 2 commits.

HeikoBecker (approved):
Thanks for your work on this! The code looks good.

LGTM.

PR #665, opened 2023-10-10
Linter + model check + folder structure fixes (2021.2)

alexanderpann:
Same PR as https://github.com/JetBrains/MPS-extensions/pull/627 but for 2021.2. We have some JARS in  `de.q60.mps.collections.libs` that collide with MPS internal JARS. I just removed the linter for the moment since this would be a breaking change.


sergej-koscejev:
Closing this because we agreed not to maintain 2021.2 any longer.

PR #670, opened 2023-10-17
Make pagination language auto-export runtime modules

HeikoBecker:
The modules are needed for editor compilation to succeed.


alamqadem:
Looks good to me, builds in MPS and from gradle ✅.

Issue #672, opened 2023-10-23
Line numbers in "MPS model Viewer" should be aligned like "Side-by-side viewer"


mgronover:
The current situation is that line numbers are hidden by the Actions placed on the same panel:

![image](https://github.com/JetBrains/MPS-extensions/assets/38247916/7b6aa0ab-dd33-4731-8bc5-e8fd05a5c83d)

- The small red rectangle shows the wrong location of the line numbers within the panel.
- The larger red rectange shows the wrong alignment of the line numbers belonging to the right editor component.

The line numbers should be placed and aligned like in the "Side-by-side viewer".

mgronover:
Fixed with https://github.com/JetBrains/MPS-extensions/pull/673

PR #673, opened 2023-10-23
Fixed alignment of line numbers when shown on right editor side

mgronover:
Issue:  #672
Line numbers in "MPS model viewer" are shown as in "Side-by side viewer"


AlexeiQ (approved):
Thx for the fix! This changes looks good, technically and functionally.
Some comments explaining the calculations would be nice and will be added.

PR #675, opened 2023-11-02
Merge maintenance/mps20213 into maintenance/mps20222

alamqadem:
Update maintenance/mps20222 branch to include latest changes from 2021.3.


sergej-koscejev:
I dunno about reviewing 347853 added lines of code... It builds so it's probably okay?

alexanderpann:
Agree. We/I agree should really merge my linter PRs. Then such a PR would be quite safe because the linters catch the bad screw-ups.

alamqadem:
> I dunno about reviewing 347853 added lines of code... It builds so it's probably okay?

Yeah, the PR was more to prove that it builds and the tests are passing.

PR #677, opened 2023-11-03
Reusable workflow for cascading merges

sergej-koscejev:
An attempt to make the cascading PR workflow reusable so that it doesn't have to be copied-and-pasted for each new version.


sergej-koscejev:
I'm going to merge it to see if it works.

PR #664, opened 2023-10-05
[Breaking] Extend grammar.wrap to check constraints properly

HeikoBecker:
Previously, if concept `A` contained concept `B` with an editor defined with grammar.rule with a grammar.wrap for `B`, constraints from `A` limiting occurrences of `B` were not taken into account when building the completion menu items. 

This PR fixes this issue.

The PR is a breaking change because it alters completion menus for all concepts that fall under the above description.

Fixes #641.

PR #679, opened 2023-11-06
Update GrammarCells constraint checks to fix empty completion menus

HeikoBecker:
The changes from #664 were causing some completion menus to become empty since the constraints were not set up to check for abstract dummy nodes. This PR addresses this issue by moving the earlier constraint test back to its original version. The filtering is then performed when building the concrete action list.


alexanderpann (changes requested):
Can you add a test case for this?

HeikoBecker:
> Can you add a test case for this?

Done in latest commit.

PR #684, opened 2023-11-07
Fix failing test in 2022.2, run migrations

sergej-koscejev:


sergej-koscejev:
Will be part of the auto merge PR

PR #688, opened 2023-11-10
Extend diagrams with a select-all plugin

HeikoBecker:
Previously, pressing Ctrl+A in a diagram label would select every node up until the including root node.

This PR adds a new plugin that fixes this issue. The key functionality was provided by @slisson, I have only fixed remaining bugs.

The last selection in the video below shows the limitations of the plugin, where only label cells are supported currently.

Peek 2023-11-10 12-24


dbinkele (approved):
LGTM

PR #689, opened 2023-11-10
Remove generator priority rule referring to nonexistent mapping

sergej-koscejev:
It causes warnings in the log during generation.


sergej-koscejev:
The mapping configuration was removed in 0ea7c5f11e91f746bb800d7bbe1e6ba18c5222ff

alexanderpann (approved):

PR #691, opened 2023-11-14
mpsutil.intentions: add a new style attribute to allow intentions in read-only-cells

alexanderpann:
The main idea is to still allow the intention menu even if nodes are, for example, just projected into the editor through a query list and read-only.

I've tried to add tests but couldn't find a way to test this. The `invoke intention` statement in the editor test doesn't check if the cell is read-only and trying to invoke a demo intention through actions or key presses also didn't work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 
 - `@NotNull` annotations in the code are now checked at run time (the `javac2` compiler is used).
 
+### Added
+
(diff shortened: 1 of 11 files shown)


HeikoBecker (changes requested):
Thanks for the work!

Overall, the changes look fine. I have one question on `StyleUtil`:

Why is there a catch-all case for read-only editor components in this function:
image
It seems to go against the idea of the PR that the style attribute regulates when intentions are allowed.

HeikoBecker (approved):
Changes look good now. LGTM.

PR #692, opened 2023-11-15
Add a new flag to pull attribute editors into edge labels

HeikoBecker:
Previously, if a model node had an attribute annotation and is used as an edge in a diagram, the attribute editor would be added as an _external box_ in the diagram (also see example node `R1` in the repo):

image

In this picture, the annotation `Component Annotation` is floating around in the diagram. Auto-layouting also does not work.

This PR adds support for pulling these annotations into the edge labels:

image

The key difference is that in the new setting, the annotation is part of the edge label and thus also automatically layouted.

To apply this change, the PR adds a new flag `pullAnnotations` as a boolean query over the node and the editor context. The standard behavior from above is retained by leaving it at the default `false` value.

This work was done jointly with @slisson.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 ### Changed
 
 - `@NotNull` annotations in the code are now checked at run time (the `javac2` compiler is used).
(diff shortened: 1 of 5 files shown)


alexanderpann (approved):
Entry in changelog is there, `editorContext` and `node` are accessible in query function and the code works in the sandbox. Looks good to me.

PR #694, opened 2023-11-17
Allow model listeners on interfaces 693

alamqadem:
This pull request adds the possibility to create model listeners on `InterfaceConcept`s.


alexanderpann:
I think that's fine. While unloading it already calls the newly compiled methods, that's why you get the errors.
Don't forget to update the changelog and the checking rule. Small refactoring possibility: It would also be nice to not log to `System.out `in the demo language but use the `log` or `message` statement.

alamqadem:
I made the changes that you suggested, if you can have a look.

I'm still getting the error if I open a project using `modellisteners`, even if I open it in the same instance alongside this `mps-extensions` branch, restarting mps doesn't solve it.
To me it's still not clear what is going on 🤔 .

alexanderpann (approved):

alexanderpann (approved):
The changes are fine. We noticed that the existing listener descriptors don't get rebuilt even when you rebuild the full language that contains the listener aspect. You have to delete the gen folders of the language first and then the rebuilding of the descriptor will succeed.

PR #697, opened 2023-11-21
fix(celllayout): additional horizontal cell was generated around optional cells

slisson:
The generator of the MPS editor language generates an additional collection cell around cells inside an alternation cell that also have a rendering condition. The problem with that is, that it has a horizontal layout which breaks surrounding indent layouts.


alamqadem (approved):
Works now for me

PR #698, opened 2023-11-22
fix(celllayout): cells where overlapping in some cases

slisson:
When cells with our own layout algorithm were mixed with the default MPS layouter then cells were overlapping especially if the content was bigger than the configured "Text width".
Now, if we replace the layouter somewhere, it is replaced in the whole subtree.


alexanderpann (approved):
I've checked if the change has any other negative impacts but couldn't find any. LGTM

PR #704, opened 2023-11-29
Add a simple tool to list dependencies used in the project in a file

HeikoBecker:
This PR adds a new action to the MPS interface under `Tools -> MPS statistics`. 

The action `Gather used languages` scans the currently opened project and records all dependencies (languages and solutions) in a file dependencies.txt in the project root. This should allow for easier dependency tracking across projects by having a point of reference.


HeikoBecker:
Thank you for the review and the improvements. LGTM.

PR #708, opened 2023-12-04
Fix size of parent cell before layouting the ports

alexanderpann:
The problem was reproducible by dropping elements from the palette. An example of broken ports can be seen here:
Screenshot 2023-12-04 at 09 17 35

The problem is that ports don't always have the correct size from the parent cell for positioning them:

Screenshot 2023-12-04 at 09 18 17

I am correcting the size of the parent cell now before layouting the ports. In the past, we just run the layouter 5 times. We still have issues with ports that are on the south side and when resizing cells with ports where the ports are at the wrong location:

Screenshot 2023-12-04 at 09 18 34


HeikoBecker (approved):
Thanks for this amazing fix! Great to have this finally solved.

LGTM

PR #709, opened 2023-12-05
Followup: port layouter fix size parent cell 2022

alexanderpann:
Small followup of https://github.com/JetBrains/MPS-extensions/pull/708 to not break manual resizing.


HeikoBecker (approved):
Thanks for the quick fix!

PR #718, opened 2023-12-12
Migrate to MPS 2023.2

alexanderpann:
This PR implements https://github.com/JetBrains/MPS-extensions/issues/699.

Remaining issues:
- <s>`com.intellij.diagnostic.PluginException: No display name specified in plugin descriptor XML file for configurable de.slisson.mps.editor.multiline.runtime.plugin.Multiline_Text_PreferencesPage`:  reported in [MPS-36601](https://youtrack.jetbrains.com/issue/MPS-36601/com.intellij.diagnostic.PluginException-No-display-name-is-specified-for-configurable-custom-preference-components)</s>
- The correct declaration of MPS.ThirdParty is again unclear in the build script. Copying the app.jar/lib.jar is not the correct solution. Can the old workaround from https://github.com/JetBrains/MPS-extensions/pull/705 be applied again or is it even necessary?


build.gradle
```diff
@@ -44,14 +44,16 @@ if (!project.hasProperty('nexusUsername')) {
 logger.info 'Repository username: {}', project.nexusUsername
 
 ext.dependencyRepositories = [
-        'https://artifacts.itemis.cloud/repository/maven-mps/'
+        'https://artifacts.itemis.cloud/repository/maven-mps/',
+//        'https://projects.itemis.de/nexus/content/repositories/mbeddr',
```
HeikoBecker, line 48: Can the two commented lines be removed?

build.gradle
```diff
@@ -139,7 +141,7 @@ configurations {
 }
 
 dependencies {
-    def modelixCoreVersion = "2.1.9"
+    def modelixCoreVersion = "1.10.3"
```
HeikoBecker, line 144: Why is the modelix version downgraded?

HeikoBecker (commented):
The changes look fine to me so far. I only found two nitpicks in the build.gradle file.

HeikoBecker:
I also do not see a `maintenance/mps20223` branch yet. It should be created before merging the PR.

alexanderpann:
> I also do not see a `maintenance/mps20223` branch yet. It should be created before merging the PR.

Yes. In the past I always created it afterwards but I guess it should be part of creating the PR .

build.gradle
alexanderpann: Probably because the migration was started with an older state of the branch.

Issue #726, opened 2023-12-18
Line Numbers: Investigate Performance Problems


alexanderpann:
For longer editors, you have a noticeable lag when they are enabled.

PR #727, opened 2023-12-18
Line numbers: Improve performance

alexanderpann:
This PR fixes https://github.com/JetBrains/MPS-extensions/issues/726

The left highlighter gets way too many relayout events which we can't influence, so we have to throttle the number of events. I've selected 250ms, which means that the line numbers only update 4 times per second or less. I've tried to create many new lines in the mbeddr documentation for testing and the update interval seems to be sufficient for small and large editors.


mgronover (changes requested):
Good solution.
- Please introduce a final variable with a name that is explaining the 250ms duration.

mgronover (approved):
LGTM

PR #730, opened 2023-12-22
editor.pagination: Implement a search bar and other improvements

alexanderpann:
Unfortunately not mergeable yet because the performance is bad for big numbers of nodes.

Screenshot 2023-12-22 at 07 59 53


HeikoBecker:
The PR has some diff also for the statistics plugin and the de.slisson.tables packages. Is this intended? They are not mentioned in the description.

alexanderpann:
> The PR has some diff also for the statistics plugin and the de.slisson.tables packages. Is this intended? They are not mentioned in the description.

I've executed Force save all at the end because some resolve infos were not up to date.

HeikoBecker:
What is the purpose of this comment: 
image

Is this getAscent relevant for the PR? The number seems somewhat arbitrary.

HeikoBecker (changes requested):
Thanks for your work on this! Once the question about the `getAscent` is answered, we can merge this.

alexanderpann:
As the comment says, the horizontal positioning of the swing component didn't work, so I had to position it manually. I couldn't figure out the right formula, so I had to hardcode the value.

HeikoBecker:
> As the comment says, the horizontal positioning of the swing component didn't work, so I had to position it manually. I couldn't figure out the right formula, so I had to hardcode the value.

Ok, thanks for the clarification. This might be a bit nitpicky but just so that future readers see the reason for the number, can you update the comment accordingly, e.g. "Hardcode horizontal position of search bar since automatic computation does not work"?

PR #732, opened 2024-01-03
Reuse cascading-merge workflow

sergej-koscejev:
Use the reusable cascading merge workflow from specificlanguages/cascading-merge.


sergej-koscejev:
GitHub says that usage of public reusable workflows needs to be enabled for the organization.

alexanderpann:
> GitHub says that usage of public reusable workflows needs to be enabled for the organization.

We already have the following checkbox enabled: Allow all actions and reusable workflows

sergej-koscejev:
Great, then this should work. I don’t see the current settings, this was just a general advice from the documentation.

alexanderpann (approved):
LGTM.

PR #744, opened 2024-01-17
Remove IDEA project files and update .gitignore

sergej-koscejev:
IDEA can import the Gradle build, no need to have its modules and libraries in Git.


sergej-koscejev:
Dear reviewers, correct me if I'm wrong but I think we don't need these files in Git.

alexanderpann (approved):
LGTM. Maybe some files were also caused when opening the Markdown documentation in IDEA.

PR #752, opened 2024-01-30
Delete leftover zero-length .msd file

sergej-koscejev:


sergej-koscejev:
Don't merge yet, there are more zero-length files I'd like to delete.

PR #754, opened 2024-02-02
diagram.runtime.model: ModelSynchronizer - avoid unnecessary calls to fromSModel_Subdiagram for sub-diagrams

danielratiu:
We are using diagrams imbricated in other diagrams (sub-diagrams), 
image

The layout information for a subdiagram is already read when the object SubdiagramECell is created
image

The ModelSynchronizer reads from SModel unnecessarily many times - this triggers many calls for imbricated diagrams.
This PR removes the reccursive refreshing of sub-diagrams from model:
image


alexanderpann (approved):
LGTM.

PR #761, opened 2024-02-09
diagram.runtime: PortLayouter handling the case when parentCell contains subdiagrams

danielratiu:


alexanderpann:
PR included cherry-picked into #760.

PR #780, opened 2024-03-01
Implement editor components with parameters

alexanderpann:
The parameter information is saved in the user object of the node instead of the root editor cell. The reason is that on first editor creation, this information can't be accessed.

For updated screenshots: see last answer

# Example

## Declaration

image

image

## Usage

image


kbirken:
Looks useful to be able to pass context information when calling the editor component. Just one proposal on syntax: It would look more consistent if the parameters were shown in the style of "overrides:" and "applicable concept:". I.e., just another section with the parameters:

```
  applicable concept:
    Interface
  parameters:
    node<> suppliedNode
    string name
```

BTW: Why is the first parameter in your example shown with type first, and the second parameter with the type after the colon?

alexanderpann:
I've implemented the feedback by @kbirken and Dennis Albrecht and also fixed the issue with the different parameter syntaxes. I like the way it looks now:

PR #790, opened 2024-03-13
MPS.Thirdparty: Fixed classloading issue due to wrong packaging 

### LINKED ISSUE #789: ThirdParty stubs: Compilation fails inside MPS and during CL build
I have a solution containing a model that uses [ArrayNode](http://127.0.0.1:63320/node?ref=39983771-4e9b-401b-a1a9-1da6c777c843%2Fjava%3Acom.fasterxml.jackson.databind.node%28MPS.ThirdParty%2F%29%2F~ArrayNode) provided via the `ThirdParty` solution. I can resolve all required references, however the compilation is failing. 

I created an example to reproduce the error. It can be found in this branch on iets3:

https://github.com/IETS3/iets3.opensource/tree/debug/ThirdPartyDependencyProblem

Here a snippet from the stacktrace:

```
 [generate] java.lang.RuntimeException: ****************************************************************************************************
   [generate] 74 errors during generation:
   [generate] Compilation problems
   [generate] package com.fasterxml.jackson.databind.node does not exist (/build/work/5fa48058a3fec642/code/languages/org.iets3.opensource/solutions/foo/source_gen/foo/bar/Fml.java:5)
   [generate] cannot find symbol
   [generate]   symbol:   class ArrayNode
   [generate]   location: class foo.bar.Fml (/build/work/5fa48058a3fec642/code/languages/org.iets3.opensource/solutions/foo/source_gen/foo/bar/Fml.java:8)
   [generate] Compilation finished: 2 errors, 0 warnings.
   [generate] Error executing target jetbrains.mps.make.facets.JavaCompile.compile
   [generate] Build failed. See previous messages for details.
   [generate] Make was not successful
   [generate] Compilation problems

```

I am using the latest artifact from `maintenance/mps20222`

arimer:
This PR fixes #789 

The changes made, are based on the  current module properties configuration and packaging of the JAXB module solution. 

**Changes are**
- MPS.Thirdparty is confiured as an intellij plugin with the id `"com.intellij"`
- stubModelEntry and modelRoot are resolved via the `${platform_lib}` path
- Packaging was done the same way it is done for [JAXB](http://127.0.0.1:63320/node?ref=r%3A874d959d-e3b4-4d04-b931-ca849af130dd%28jetbrains.mps.ide.build%29%2F7972607346140852471)

**Simple Test**
- in addition I've added a simple test scenario
- We now have solution that uses some BL classes provided by MPS.Thirdparty
- This solution is part of the test build conflig and should be generated and compiled during CL build
- in case we have classloading issues, its compilation will fail

### DIFF
--- a/code/.mps/modules.xml
+++ b/code/.mps/modules.xml
@@ -192,6 +192,7 @@
       <modulePath path="$PROJECT_DIR$/tables/solutions/de.slisson.mps.testutils/de.slisson.mps.testutils.msd" folder="tables" />
       <modulePath path="$PROJECT_DIR$/tables/solutions/test.de.slisson.mps.tables/test.de.slisson.mps.tables.msd" folder="tables" />
       <modulePath path="$PROJECT_DIR$/third-party/solutions/MPS.ThirdParty/MPS.ThirdParty.msd" folder="3rd-party" />
+      <modulePath path="$PROJECT_DIR$/third-party/solutions/third.party.usage.test/third.party.usage.test.msd" folder="" />
       <modulePath path="$PROJECT_DIR$/tooltips/solutions/de.itemis.mps.tooltips.runtime/de.itemis.mps.tooltips.runtime.msd" folder="tooltips" />
       <modulePath path="$PROJECT_DIR$/treenotation/com.mbeddr.mpsutil.treenotation.runtime/com.mbeddr.mpsutil.treenotation.runtime.msd" folder="treenotation" />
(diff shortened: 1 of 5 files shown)


alexanderpann (changes requested):
It looks good to me, just one small change: Can you not only import something from the stub in your test solution but also reference a class itself? We want to check if it compiles.

arimer:
Dang! I though I added it, but forgot it :/ 

Now it should be done

alexanderpann (approved):
LGTM. I assume you have tested this change in another project with this MPS-Extensions dependency. The build script of this second project also shouldn't complain.

arimer:
I did try it on the branch mentioned here https://github.com/JetBrains/MPS-extensions/issues/789. However I did not take the artifacts from nexus but use the local ones.

PR #817, opened 2024-04-05
Querylist: support model checking

alexanderpann:
Open question: 
- Could[ QueryListChecker#needsUpdate](http://127.0.0.1:63320/node?ref=r%3A5f74240b-ce7f-4ae5-99bc-bf1a210bda6d%28com.mbeddr.mpsutil.editor.querylist.runtime.plugin%29%2F4268492325946535452) be implemented and if so, how?
- Is enabling the model checker possible for nodes without a model? (my guess: no)

The feature can be tested [here](http://127.0.0.1:63320/node?ref=c58b797b-c53c-4442-abff-4d54c72169ea%2Fr%3A5f236bd8-eb21-40b6-825a-9e614ca27175%28com.mbeddr.mpsutil.editor.querylist.sandbox%2Fcom.mbeddr.mpsutil.editor.querylist.sandbox%29%2F3813940761427382564).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 ### Fixed
 
 - com.mbeddr.mpsutil.projectview: Class reloading of project views now works.
+- com.mbeddr.mpsutil.editor.querylist: Query lists now support model checking for non-dynamically generated nodes and `collapse by default` is generated correctly.
+- de.slisson.mps.reflection: To fix the compilatation issues, the language is now generated earlier in the generation plan.
 
 ## March 2024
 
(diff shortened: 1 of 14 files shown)


slisson:
The implementation of `needsUpdate` can be as simple as in [AbstractTypesystemEditorChecker](https://github.com/JetBrains/MPS/blob/0086c63d6c94d1bbeee860dab17b7b1aebcf70d9/editor/typesystemIntegration/source/jetbrains/mps/typesystem/checking/AbstractTypesystemEditorChecker.java#L99) and at least that's required. Otherwise, MPS consumes a lot of unnecessary CPU time.

The implementation of `QueryListChecker.update` is very inefficient. For each node it checks the whole model and then filters the messages for that node. Running the checkers only once per model will already improve the performance a lot and shouldn't be much effort. It should also be possible to check only root nodes instead of the whole model.

alexanderpann:
*   `needsUpdate` is fixed.
*   `QueryListChecker.update` is now less inefficient.
*   The feature can now be disabled globally.

> Nodes that are not attached to a model shouldn't even be shown in the editor.

I enabled this feature a while ago for read-only editors where it is supported. I guess there is still no way that we can make it work with model checking.

slisson:
I optimized the performance a bit more.
Also added actions so that users can disable/enable the model checker themselves.

PR #830, opened 2024-04-24
Migration tasks

sergej-koscejev:
* Add Gradle tasks to run migrations from command line.
* Build sandbox and demo languages as part of the build (necessary in order to be able to run migrations).
* Run migrations once.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,5 @@ code/model-api/org.modelix.model.api/lib
 code/dependencies.txt
 .idea/
 *.iml
+
+TEST*.xml
--- a/build.gradle
+++ b/build.gradle
@@ -1,11 +1,14 @@
 plugins {
-    id 'de.itemis.mps.gradle.common' version '1.13.+'
+    id 'de.itemis.mps.gradle.common' version '1.24.0.+'
     id "com.github.breadmoirai.github-release" version "2.4.1"
     id 'maven-publish'
     id 'base'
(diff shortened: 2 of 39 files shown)


sergej-koscejev:
After this PR is propagated to master, we could modify the TeamCity build to verify that everything stays migrated, similar to [how it's done in mps-qa](https://github.com/mbeddr/mps-qa/blob/7f48e9ef9125300d591d3773868a2746e9fdc862/.github/workflows/build_mpsqa.yml#L49).

alexanderpann (changes requested):
By running the migrations, you applied the migrations to the migration tests in `test.com.mbeddr.mpsutil.blutil.doc@tests`which we want to test.

The rest of the changes look good to me.

sergej-koscejev:
I'll need to discuss this with JetBrains. I can't ignore a single model for the purposes of migrations. I could ignore the whole module with tests but that would mean that other migrations (such as those from base language or the test language) also would not apply to it.

Alternatively, I could ignore just one specific migration on one specific module but I'm not sure it's a good solution. Somebody could still migrate the code by accident by re-running migrations from MPS.

sergej-koscejev:
JetBrains don't have any solution for this. Looks like the best approach would be to modify the migration so that it ignores migration tests. What do you think @alexanderpann?

alexanderpann:
Fine by me.

PR #833, opened 2024-04-26
blutil: avoid migrating input/output nodes of MigrationTestCase

sergej-koscejev:
The implementation uses magic constants for the MigrationTestCase input/outputNodes containment links to avoid adding a dependency on jetbrains.mps.lang.test language.


alexanderpann (approved):
LGTM.

PR #886, opened 2024-07-12
editor.diagram: Implement node/edge/label ELK options

alexanderpann:
Most of the remaining [ELK layout options](https://eclipse.dev/elk/reference/options.html) should be covered. The old options can be set in the inspector of the diagram cell. The edge options can be set in the inspector of the edge cell and the label options can be set in the label cell. The options should also work in the generic box query and generic edge query.

For the elk sandbox, there are stylesheets (e.g. [LayeredTestStyleSheet](http://127.0.0.1:63320/node?ref=r%3A37620960-62b9-4f89-b6a6-51b609691f93%28de.itemis.mps.editor.diagram.demo.elk.editor%29%2F5080030193841786852&project=mps-extensions)) that can be enabled through the checkbox in the sandbox itself. One style sheet item (inline edge labels) is also [enabled](http://127.0.0.1:63320/node?ref=r%3Ae0c32a3c-18a4-4f21-8a60-d16a394a9606%28de.itemis.mps.editor.diagram.demoentities.editor%29%2F3462102746003563044&project=mps-extensions) in the demoentities language. When it works, it draws the labels on top of the edges instead of next to them.

The PR also contains some other small fixes.


HeikoBecker (approved):
Code changes look fine. Thanks a lot for the work!

PR #893, opened 2024-07-26
Use mps.make task for building

sergej-koscejev:


sergej-koscejev:
This should make builds faster but currently tests don't run.

sergej-koscejev:
Indefinitely postponed due to other priorities.

PR #915, opened 2024-08-07
made querylist model checking result propagation more responsive to cancellation and added optional timeout

enikao:


alexanderpann (approved):
Thanks for the improvement. LGTM

PR #916, opened 2024-08-07
keep typesystem checker computation mode alive inside querylist

enikao:
Found another performance issue with querylist model checking: It doesn't preserve the `TypeCheckingContext.NonTypesystemComputationMode`. If we have (expensive) checks marked as `do not apply on the fly: true`, they were still executed by querylist.

Unfortunately, I didn't find a better way than to duplicate most of `NonTypesystemChecker.getErrors()`.


enikao:
In `ComputationModePreservingNonTypesystemChecker.getErrors()`: I'm not sure whether we should reset computation mode in `finally`. What do you think?

alexanderpann (approved):
I see that in LanguageEditorChecker it is also set back to "off".  It doesn't seem to harm but I also can't judge because I don't know enough about the type system.

PR #923, opened 2024-08-13
editor.diagram: add a new option to allow boxes to set a new required minimum size

alexanderpann:
The idea is that you want to sometimes shrink a diagram box to a size below the required size, so that the automatic line wrapping kicks in e.g. with Text from the richtext language. I set the max-width stylesheet item for that. Fit size all and autolayout should respect this setting. @slisson would it be possible to somehow also do this for sizes that are above the minimum size?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is *loosely* based on [Keep a Changelog](https://keepachangelog.com/e
 
 ### Added
 
+- *de.itemis.mps.editor.diagram*: A new option for diagrams was added to allow boxes to set a new required minimum size.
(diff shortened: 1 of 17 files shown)


slisson:
Maybe we should add the methods `LayoutEngine.getPreferredSize(EditorCell cell, Size sizeConstraint)` and `LayoutEngine.getMinSize(EditorCell cell, Size sizeConstraint)` and use them inside `MPSCell.getMinSize` to calculate a reasonable width by default without any configuration.

One option could be to decide base on the ratio between width and height. If the cell is very wide but has a small height we could call getPreferredSize with a constraint for the width and aim for a more visually pleasing ratio.

alexanderpann:
I've tried the suggested alternative approach, and it didn't work out at all. The current solution is also not ideal, but I'll merge because it at least improves the situation.

PR #926, opened 2024-08-14
Fix EditorComponentCreationListener uses

artem-tikhomirov:
Since 2024.1, MPS EditorExtensionRegistry become a CoreComponent (used to be IDEA ProjectComponent).
To access both IDEA and MPS components, use MPSProject.getComponent() API.
Note, this approach is necessary for compatibility between MPS 2024.1 and earlier versions as long as same MPS-extensions code is in use both in 2023.x and 24.x versions. Once there's 24.1+ only (23.3 forks to maintenance), `ProjectPlugin` parts contributing EditorExtension shall get converted into `ApplicationPlugin`, as EditorExtensionRegistry CoreComponent is application-wide (unlike IDEA's ProjectComponent it used to be up and including to 2023.3).


alexanderpann (approved):
LGTM. Is there a reason to set the `EditorExtension` instance to null in the dispose block?

PR #929, opened 2024-08-22
mps.tables: implement sticky cells

alexanderpann:
Sticky table cells were added. Add the style sheet item `horizontal-sticky-cell` or `vertical-sticky-cell` to a table cell to activate the feature. A sticky cell will be rendered on the left or at the top of the editor when it would not be visible otherwise when the users scrolls up or down the table:

Screenshot 2024-08-22 at 07 47 54

The only known bug is that when you use horizonal and vertical sticky cells at the same time, they will overlay. The cells are also only painted, so you can't interact with them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ code/dependencies.txt
 *.iml
 
 TEST*.xml
+
+# screenshots
+code/cell-screenshots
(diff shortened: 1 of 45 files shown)


HeikoBecker (approved):
Thanks a lot for the work! The feature looks very neat.

PR #949, opened 2024-09-16
editor.diagram: add support for grid snapping

alexanderpann:
Diagrams can now show a grid that is used for snapping elements to it. It can be configured (visibility, grid snapping, grid color/size/style) in the editor definition of the diagram itself:

Screenshot 2024-09-16 at 09 20 29

The elements sometimes don't snap to the grid itself but that seems to be an issue in the library itself. Taken from https://github.com/jgraph/mxgraph/blob/ff141aab158417bd866e2dfebd06c61d40773cd2/java/src/com/mxgraph/swing/handler/mxGraphHandler.java#L1105: 

> Drag image _size_ depends on the initial position and may sometimes
> not align with the grid when dragging. This is because the rounding of the width
> and height at the initial position may be different than that at the current
> position as the left and bottom side of the shape must align to the grid lines.
>  Only fix is a full repaint of the drag cells at each new mouse location.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is *loosely* based on [Keep a Changelog](https://keepachangelog.com/e
 
 ### Added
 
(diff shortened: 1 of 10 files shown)


HeikoBecker (approved):
Thanks for implementing this feature. Looks great!

LGTM.

PR #954, opened 2024-09-19
mps.tables: make the dummy link work in the cell inspector

arimer:
Backporting bugfix
(cherry picked from commit 81a2d7acce468dfb2af40c3e3e191cf6ecffdf38)


arimer:
Interessting, our TC configuration seem to find some pending migrations. Would be a cool feature if we could forward that info back to github or so.

arimer:
Hmmm If I open the project with MPS 2022.2.3 and run the migratino assistant no changes are found 🤔

sergej-koscejev:
It found some resolveInfo updates. The simplest way to fix it is to run `gradlew migrate remigrate`, then check and commit the result.

alexanderpann (approved):

Issue #958, opened 2024-09-24
HeaderGridFactory does not (always) take HeaderReference into account


arimer:
When initialising a HeaderGrid via [HeaderGridFactory.createFromObjectList()](http://127.0.0.1:63320/node?ref=r%3Aea653f2d-c829-4182-b311-a544ef1f4c1f%28de.slisson.mps.tables.runtime.gridmodel%29%2F1397920687874776980)  the `HeaderReference` is ignored. The returned HeaderGrid  instance and its `myHeaderStore` is not reflecting the initially passed in HeaderReference. 

Therefore it is not possible to retrieve the header from the store via [HeaderStore.getHeader(HeaderRef)](http://127.0.0.1:63320/node?ref=r%3Aea653f2d-c829-4182-b311-a544ef1f4c1f%28de.slisson.mps.tables.runtime.gridmodel%29%2F8791219374217122041).

MPS 2022.2

Issue #955, opened 2024-09-19
EndRowCells: Delete Action always deletes the whole table


arimer:
Problem accours for MPS 2022.2

### The problem
In case you press Delete/Backsspace Key on a EndRowCell, the entire Table is about to be deleted.

<img width="345" alt="grafik" src="https://github.com/user-attachments/assets/711b3224-da7a-496a-a989-30806482eada">

### Expected Behaviour
Pressing Delete/Backsspace Key on a EndRowCell always proposes the deletion of the row in question.

Issue #953, opened 2024-09-19
RowEndCell: Is not generated for tables that use a Table.rowHeaders


arimer:
In case a Table-Editor is defining any [Table.rowHeaders](http://127.0.0.1:63320/node?ref=r%3A9805308e-34d7-4576-b7d2-a758b4bd38a3%28de.slisson.mps.tables.structure%29%2F1397920687864865685) the RowEndCell is not generated.

PR #974, opened 2024-10-09
Tables: Various fixes to table end cells in combination with insert and deletion handler

alexanderpann:
@arimer to test the changes, please have a brief look at the MPS-Extensions [Tables Sandbox](http://127.0.0.1:63320/node?ref=r%3A7b909ba9-e58e-4fbd-8829-ecfbd89930fa%28de.slisson.mps.tables.sandbox%29%2F1397920687866914332) to check for broken stuff. I didn't add any delete or insert handlers there.

For proper testing, you can open https://github.com/IETS3/iets3.opensource/tree/bugfix/tables as a second project and check instances of the following concepts: DataTable, Sheet, LookupTable, DecTab.

When the default implementation for deletion is used (not derived from a header deletion handler), there is a differentiation between BACKSPACE (delete to the left of cursor) and DELETE (delete to the right of the cursor). For the derived delete actions we could create a followup ticket. For INSERT (enter) and INSERT_BEFORE (Shift + Enter) the same applies. For all actions this means that depending on the ACTION it modifies the row before the current row, the current row or the next row.  The left and right cell therefore modify different rows. Example:
DELETE on a left end cell deletes the current row since it is to the right of the cell, DELETE on a right end cell deletes the next row.

I didn't open a PR in IETS3 yet because without this PR the other PR doesn't build. I also added a few helper methods such as `Grid#setEndCellAction` that are unused in both repos as the moment but could be useful for other people.

- fixes https://github.com/JetBrains/MPS-extensions/issues/953
- fixes https://github.com/JetBrains/MPS-extensions/issues/955 in combination with https://github.com/IETS3/iets3.opensource/tree/bugfix/tables
- fixes https://github.com/JetBrains/MPS-extensions/issues/958

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,6 @@ All notable changes to this project are documented in this file.
 
 The format is *loosely* based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) .The project does *not* follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
-## October 2024
-
-### Fixed
-
-- *de.slisson.mps.tables*: Various fixes to table end cells in combination with insert and deletion handlers were made.
-
 ## August 2024
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project are documented in this file.
 
 The format is *loosely* based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) .The project does *not* follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## October 2024
+
+### Fixed
+
+- *de.slisson.mps.tables*: Various fixes to table end cells in combination with insert and deletion handlers were made.
+
 ## August 2024
 
 ### Fixed
--- a/code/solutions/de.itemis.mps.extensions.changelog/models/de.itemis.mps.extensions.changelog.mps
+++ b/code/solutions/de.itemis.mps.extensions.changelog/models/de.itemis.mps.extensions.changelog.mps
@@ -227,6 +227,76 @@
         </node>
       </node>
     </node>
+    <node concept="15bmVD" id="1ieETfbmCxq" role="15bmVC">
+      <node concept="15ShDW" id="1ieETfbmCxn" role="15bq2Y">
+        <property role="15ShDY" value="Po4Z58IgBa/October" />
+        <property role="15ShDw" value="2024" />
+      </node>
+      <node concept="15bAme" id="1ieETfbmCxo" role="15bAlL">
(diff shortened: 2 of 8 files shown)


arimer (changes requested):
Awesome refactoring 👍🏼 
Just some minor findings we should address before merging: 

- lets surround the whole logic inside the for loop with that[ header!=null](http://127.0.0.1:63320/node?ref=r%3A9eebd85f-0264-4bca-b8dc-0e1b71e196f7%28de.slisson.mps.tables.generator.template.main%40generator%29%2F244222589689910240) check
- Inside the ActionSettingAdapter, [this action](http://127.0.0.1:63320/node?ref=r%3Aea653f2d-c829-4182-b311-a544ef1f4c1f%28de.slisson.mps.tables.runtime.gridmodel%29%2F2399441400661704708) is not matching properly the case condition. I guess it should be a `setDeleteAction()`
- We are under `rightEnabled`. How about we rename [this variable](http://127.0.0.1:63320/node?ref=r%3Aea653f2d-c829-4182-b311-a544ef1f4c1f%28de.slisson.mps.tables.runtime.gridmodel%29%2F8156363746484502838)to something "right"
- This [HeaderReference key might](http://127.0.0.1:63320/node?ref=r%3Aea653f2d-c829-4182-b311-a544ef1f4c1f%28de.slisson.mps.tables.runtime.gridmodel%29%2F8271849212530108458) be not stable enought. Should we change the [getFallBackText()](http://127.0.0.1:63320/node?ref=r%3Aea653f2d-c829-4182-b311-a544ef1f4c1f%28de.slisson.mps.tables.runtime.gridmodel%29%2F1381472365344099779)to also take the `value` into account?

arimer:
The usecases in iest3 tables work as epected 💯 

However I gave up in understanding why a lookup table looks the way it looks like: 
grafik

I don't know why  there is the 2 literal at all and why the end-cell is not show there, but this should not hinder us in merging.

alexanderpann:
> How about we rename [this variable](http://127.0.0.1:63320/node?ref=r%3Aea653f2d-c829-4182-b311-a544ef1f4c1f%28de.slisson.mps.tables.runtime.gridmodel%29%2F8156363746484502838)to something "right"
We still query the left create handler, so the name is fine.

> hould we change the [getFallBackText()](http://127.0.0.1:63320/node?ref=r%3Aea653f2d-c829-4182-b311-a544ef1f4c1f%28de.slisson.mps.tables.runtime.gridmodel%29%2F1381472365344099779)to also take the value into account?

I changed `SNodeHeaderReference#getFallBackText` to match `StringHeaderReference#getFallBackText`.

> I don't know why there is the 2 literal at all and why the end-cell is not show there, but this should not hinder us in merging.

Agree but you can't insert a row here anyway, so it is correct.

arimer (approved):
Thanks for the cleanup 👍🏼 

Changes look fine now.

PR #1000, opened 2024-10-31
rerouted all dependencies on guava collections from MPS.IDEA to de.q60.mps.collections.libs

enikao:
rerouted all dependencies on com.google.common.collect(MPS.IDEA/) to com.google.common.collect(de.q60.mps.collections.libs/)


enikao:
I'd imagine the most critical change is in the Modelix adapter.
@slisson How to test whether that still works?

alexanderpann:
Modelix model-api is deprecated and will be removed in the future as far as I remember, so it is not critical.

alexanderpann (approved):
LGTM, thanks for the cleanup.

enikao:
@alexanderpann Thanks for merging

PR #1005, opened 2024-11-05
Migrate to MPS 2024.1.1

alexanderpann:
- fixes https://github.com/JetBrains/MPS-extensions/issues/918

Nothing special changed. I only had issues with icon loading for plugin models that are nested in other models. Those models couldn't be opened in the logical view. The workaround seems to be to set `Generate initializer config` to true in the `StandalonePluginDescriptor`. The changes related to introducing the XML stubs are related to mbeddr. We need them available in MPS-Extensions since we depend in the batik stubs on then.

Note: We skipped MPS 2023.3 on purpose and decided to immediately go to MPS 2024.1.

Encountered known MPS issues:

- https://youtrack.jetbrains.com/issue/MPS-36888/java.lang.UnsupportedOperationException-not-implemented
- https://youtrack.jetbrains.com/issues/MPS?q=java.lang.IllegalStateException:%20Write%20actions%20are%20prohibited
- https://youtrack.jetbrains.com/issue/MPS-37823/Invalidate-caches-and-restart-Failed-to-close-and-dispose-all-projects
- https://youtrack.jetbrains.com/issue/MPS-37533/idea.log-DisposalException-on-MPS-shutdown 

Other non-migration-related issues:

- https://github.com/JetBrains/MPS-extensions/issues/1003
- https://github.com/JetBrains/MPS-extensions/issues/1004

PR #1069, opened 2024-12-03
xmodelgen: use enhanced transform in mapping configurations step

wsafonov:
Transform.[languages](http://127.0.0.1:63320/node?ref=r%3A4a23ef0d-9c2f-48a6-8597-fbdd5b11f792%28jetbrains.mps.lang.generator.plan.structure%29%2F2944629966652439181) has been deprecated since 2023.1 (see [EnhancedTransformStep](http://127.0.0.1:63320/node?ref=r%3A18ce53c3-75ab-44c5-9292-2f157dd0081e%28jetbrains.mps.lang.generator.plan.migration%29%2F5914941733269717090)). [MappingConfigStep](http://127.0.0.1:63320/node?ref=r%3Ad1800018-44fb-4b2e-b3ae-2afea554d27b%28de.itemis.mps.hacks.xmodelgen.structure%29%2F961590472824346305) should be updated and existing instances migrated to use the new, enhanced version of the Transform.

### DIFF
--- a/code/hacks/languages/de.itemis.mps.hacks.xmodelgen/de.itemis.mps.hacks.xmodelgen.mpl
+++ b/code/hacks/languages/de.itemis.mps.hacks.xmodelgen/de.itemis.mps.hacks.xmodelgen.mpl
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<language namespace="de.itemis.mps.hacks.xmodelgen" uuid="c5eeb6dc-2f3d-45ae-a7be-929daeb6bda1" languageVersion="0" moduleVersion="0">
+<language namespace="de.itemis.mps.hacks.xmodelgen" uuid="c5eeb6dc-2f3d-45ae-a7be-929daeb6bda1" languageVersion="1" moduleVersion="0">
   <models>
     <modelRoot contentPath="${module}" type="default">
       <sourceRoot location="models" />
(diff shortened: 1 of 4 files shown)


alexanderpann:
FYI: I completely removed the language in https://github.com/JetBrains/MPS-extensions/pull/1005 for 2024.1.
There were references to generator configurations that were not working at all for a long time. It was unfixable. Is there a reason we want to introduce the language again?

wsafonov:
> FYI: I completely removed the language in #1005 for 2024.1. There were references to generator configurations that were not working at all for a long time. It was unfixable. Is there a reason we want to introduce the language again?

What about existing usages in customer projects? Some of them might need a bigger refactoring effort in order to switch to the standard language. GIven that 2024.1 might be the next version to migrate to, it could be not enough enough time for the customer projects to find a proper replacement solution. So there is a risk of still being dependent on the existing language in 2024.1.

alexanderpann:
This language was mainly about referencing mapping configurations directly in the gen plan and it used to be used by an action in mbeddr to derive a gen plan (which is now a built-in action).
Comment by @slisson : 
> I wouldn't use that language anymore. Each mapping configuration must be moved to its own generator module ... You can run individual generators in a GenPlan, but I don't know how well that works. Features that are rarely used often contain bugs.

I actually tried to fix the language, but I really had to removed it since it was not compatible with current MPS versions anymore and showed error messages that were unfixable. The last time this language was touched was in 2019.

wsafonov:
> We still have an important use case in the customer project for the documentation generation that is relying on the mapping config steps. I would also like to move away from this, but I'm not sure if this refactoring would be an easy task and can be finished before the migration to MPS 2024.1, as it would probably require touching the mbeddr doc generators.
> 
> > I actually tried to fix the language, but I really had to removed it since it was not compatible with current MPS versions anymore and showed error messages that were unfixable. The last time this language was touched was in 2019.
> 
> So in other words this means that starting from MPS 2024.1 this language has definitely become not usable anymore, right? This is important to know since we need then to start preparing and pushing the migration for the customer use cases.

wsafonov:
Based on the evaluation of customer use cases still relying on `MappingConfigStep`, supporting the language `de.itemis.mps.hacks.xmodelgen` doesn't appear necessary anymore - we expect that a couple of remaining non-critical use cases can be successfully refactored to use the standard, improved genplan language. Closing this PR.

PR #1075, opened 2024-12-04
include a GitHub workflow for vulnerability scanning

alexanderpann:
Currently, the build won't fail.  We would have to add an argument like --failOnCVSS 7 to fail the build for issues with high severity. For example, for jgraphx we would have to add it as a false positive since we can't update the library.
You can find the output for a run on the actions page: https://github.com/JetBrains/MPS-extensions/actions/runs/12163684040?pr=1075. There is an artifact [Depcheck report](https://github.com/JetBrains/MPS-extensions/actions/runs/12163684040/artifacts/2274387072).
Alternativelly, we could only add it for the `workflow_dispatch` event. Then you would have to call it manually.

### DIFF
--- a/.github/workflows/vulnerability-scanning.yml
+++ b/.github/workflows/vulnerability-scanning.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
       - name: Call setup
-        run: ./gradlew build_allScripts
+        run: ./gradlew dependencies
       - name: Depcheck
         uses: dependency-check/Dependency-Check_Action@main
         env:
--- a/.github/workflows/vulnerability-scanning.yml
+++ b/.github/workflows/vulnerability-scanning.yml
@@ -1,0 +1,33 @@
+on: [workflow_dispatch]
+
+jobs:
+  depchecktest:
+    runs-on: ubuntu-latest
+    name: depecheck_test
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 17
+      - name: Setup Gradle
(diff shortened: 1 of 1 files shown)


.github/workflows/vulnerability-scanning.yml
```diff
@@ -0,0 +1,35 @@
...
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 17
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
+      - name: Call setup
+        run: ./gradlew dependencies
+      - name: Depcheck
+        uses: dependency-check/Dependency-Check_Action@main
```
sergej-koscejev, line 20: I don't think it's a good idea to use a branch like that, better pin to a commit and have Renovate update it regularly.

.github/workflows/vulnerability-scanning.yml
```diff
@@ -0,0 +1,35 @@
...
+        env:
+          # actions/setup-java changes JAVA_HOME, so it needs to be reset to match the depcheck image
+          JAVA_HOME: /opt/jdk
+        id: Depcheck
+        with:
+          project: 'MPS-extensions'
+          format: 'HTML'
+          out: 'reports'
+          args: >
+            --exclude github/workspace/build/mps/**
```
sergej-koscejev, line 30: should this use `$GITHUB_WORKSPACE` or `{{ github.workspace }}` perhaps?

.github/workflows/vulnerability-scanning.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Upload Test results
+        uses: actions/upload-artifact@master
```
sergej-koscejev, line 32: Same comment about pinning to a particular tag or commit.

sergej-koscejev (approved):

PR #1079, opened 2024-12-05
Reduce errors to warnings when finding missing policies.

mfsoliveira:
For large languages or prototyping it is difficult to define and test merge policies for all concepts at the same time.
After some experiments trying to create default policies, we saw that we could not avoid the definition of merge policy at least for elements found the model, because the identity function must be defined.

As workaround I just reduce errors to warnings when finding missing policies. Additionally I added a flag to indicate an user is interesting to use a partial merge policy, so that some assertions and checks during the merge could be ignored.


mfsoliveira:
I messed up branch from the master, change code that need to merge in 2023.2. 
I branched again, now from correct 2023.2 and picked up the commits I need. The PR is #1083.
I am closing this PR

PR #1125, opened 2025-01-15
Make intentions executable even if they have children, just like in MPS IntentionMenuProducer.

malueck:
Based on this line: https://github.com/JetBrains/MPS/blob/master/editor/intentions-runtime/source/jetbrains/mps/editor/intentions/IntentionMenuProducer.java#L152

Otherwise an action that is a popup group is not "performable" in the ordinary sense.

Only affects versions for MPS 2023.2 and higher which have the MyIntentionMenuProducer implementation.


alexanderpann (approved):
LGTM, thanks for the fix! I checked with 2023.2 and 2024.1 and it works. btw,2023.2 was not affected by the original issue. There must have been some change in the IntelliJ platform in 2024.1.

PR #1151, opened 2025-02-04
Migrate compare language from mbeddr to mps-extention

### LINKED ISSUE #30: Migrate com.mbeddr.mpsutil.compare
Migrate com.mbeddr.mpsutil.compare

Url: http://127.0.0.1:63320/node?ref=r%3A742f344d-4dc4-4862-992c-4bc94b094870%28com.mbeddr.mpsutil.dev.build%29%2F745179663200110966&project=com.mbeddr.build

see migration guide at: https://jetbrains.github.io/MPS-extensions/Migrating/

mfsoliveira:
This PR migrates the compare language and its dependencies from mbeddr to mps-extensions.

The Demo works fine from IDE but it breaks in the build. 
There is some special test in the demo that show cases the use of compare with generator output. It was not working well in mbeddr, thus it was not introduced in the build-test script. Because the migration task in CI, requires all existing languages and solution to be defined in the build script, I decided to remove the show-case completely. 
  
fixes #30

### DIFF
--- a/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
+++ b/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
@@ -16528,6 +16528,11 @@
         <property role="2iUeEt" value="itemis AG" />
         <property role="2iUeEu" value="https://www.itemis.com/en/it-services/methods-and-tools/mps" />
       </node>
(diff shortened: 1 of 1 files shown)


alexanderpann (approved):
Thanks for migrating the language to MPS-Extensions!

LGTM, just one small addition: Can you please add a description to the newly created idea plugin?

PR #1152, opened 2025-02-04
stubs: use external compilation in stubs solutions to enable MPS module classloading

wsafonov:
Enable external compilation for `de.itemis.stubs.batik` and `de.itemis.stubs.xml` stubs solutions to fix classloading issues with deployed modules.

- Stubs solutions that contain classes not compiled by MPS should be explicitly configured to be compiled externally according to the [Migration Guide](https://www.jetbrains.com/help/mps/migration-guide.html#mps20241_migration). Otherwise, such solutions don't use MPS module classloader during runtime and specifically don't consider module dependencies for classloading. This gets particularly configured by the generated build script that contains the generated module descriptor for a stubs solution. If external compilation is specified for the module, its module dependencies are also added to the generated descriptor and can be used in runtime classloading:

  image
   
- The PR also disables extension contribution for both stubs solutions, since neither of them provides any extensions to MPS (such as plugins or extension points).

- Additionally, issue navigator config for the Github project has been added to simplify navigation to PRs and issues.

### DIFF
--- a/code/.mps/vcs.xml
+++ b/code/.mps/vcs.xml
@@ -1,5 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="IssueNavigationConfiguration">
+    <option name="links">
+      <list>
+        <IssueNavigationLink>
+          <option name="issueRegexp" value="#(\d+)" />
+          <option name="linkRegexp" value="https://github.com/JetBrains/MPS-extensions/pull/$1" />
+        </IssueNavigationLink>
+      </list>
+    </option>
+  </component>
   <component name="VcsDirectoryMappings">
     <mapping directory="$PROJECT_DIR$/.." vcs="Git" />
   </component>
--- a/code/batik/solutions/de.itemis.stubs.batik.msd
+++ b/code/batik/solutions/de.itemis.stubs.batik.msd
@@ -7,7 +7,7 @@
     </modelRoot>
   </models>
   <facets>
-    <facet type="java" compile="off" classes="mps" ext="yes">
+    <facet type="java" compile="ext" classes="mps" ext="no">
       <classes generated="true" />
       <library location="${module}/lib/xmlgraphics-commons.jar" />
       <library location="${module}/lib/batik-all.jar" />
--- a/code/xml/solutions/de.itemis.stubs.xml.msd
+++ b/code/xml/solutions/de.itemis.stubs.xml.msd
@@ -7,7 +7,7 @@
     </modelRoot>
   </models>
   <facets>
-    <facet type="java" compile="off" classes="mps" ext="yes">
+    <facet type="java" compile="ext" classes="mps" ext="no">
       <classes generated="true" />
       <library location="${module}/lib/xercesImpl.jar" />
       <library location="${module}/lib/xml-apis-ext.jar" />


alexanderpann:
I don't understand why you want to disable extension contribution. Just have a look at the MPS documentation where it is a) mentioned that you set it and b) I know at least from MPS 2023.2 that not setting it leads to classloading issues: https://www.jetbrains.com/help/mps/getting-the-dependencies-right.html#addingexternaljavaclassesandjarstoaproject-runtimesolutions. The other change looks correct to me.

wsafonov:
> I don't understand why you want to disable extension contribution. Just have a look at the MPS documentation where it is a) mentioned that you set it and b) I know at least from MPS 2023.2 that not setting it leads to classloading issues: https://www.jetbrains.com/help/mps/getting-the-dependencies-right.html#addingexternaljavaclassesandjarstoaproject-runtimesolutions. The other change looks correct to me.

Well, because according to the dialog description setting it to Yes would not really make sense and is confusing - the dialog says that it's required only if the module provides extensions to MPS which is definitely not the case with the external java libs. On the other side, the documentation says setting it to Yes is needed to have MPS reload the classes properly. Reloading of classes from external jars is not really relevant, since they are normally not getting changed continuously. It could be only become relevant in development environment, when working on mps-extensions source. But even then the library and the classes are getting perfectly loaded and unloaded when changing the module descriptor or even updating the jar on the disk. Anyway, the most important use case when working with such runtime libraries in deployment context works without any issues (tested in a customer project context using GraphViz visualisation requiring Batik's SVG documents). So I don't see, where this should be a problem.

But if you remember what kind of classloading issues you observed, we could check that as well. Otherwise I would suggest, that starting from MPS 2024.1 we leave it configured like that, according to the most logical choice. And if we run into any issues again, we could always change it back. And report it then to JB based on the specific scenario where it wouldn't work. We could even report a ticket for now to clarify why the documentation is recommending something different than what's suggested by the UI - to get rid of this inconsistency.

alexanderpann:
The stubs in MPS itself have those settings set to External Compiler, Module Contributor and Contributes extensions to MPS is set to no. Ok, let's try the change. The last time we had issues in mbeddr.core (https://github.com/mbeddr/mbeddr.core/pull/2444/files) that classes from org.xml could not be found and for one of the the migration I had to change this setting for all stubs. This was my ticket and they didn't comment on it: https://youtrack.jetbrains.com/issue/MPS-36814/Update-documentation-about-solution-settings-for-stubs

wsafonov:
> The stubs in MPS itself have those settings set to External Compiler, Module Contributor and Contributes extensions to MPS is set to no. Ok, let's try the change. The last time we had issues in mbeddr.core (https://github.com/mbeddr/mbeddr.core/pull/2444/files) that classes from org.xml could not be found and for the migration I had to change this setting for all stubs. This was my ticket and they didn't comment on it: https://youtrack.jetbrains.com/issue/MPS-36814/Update-documentation-about-solution-settings-for-stubs

Thanks for providing the PR links! I see that in that mbeddr PR you have changed the classloader in org.xml solution from `no` to `mps` - maybe that has made the difference and not actually the setting for extensions contribution. At least this would make more sense to me. But let's proceed with this setup for now and observe this in the field if anyone would run into any troubles with this.

Btw, since we have more pure stubs solutions, like `org.apache.commons` and `de.q60.mps.collections.libs`, I would suggest for consistency reasons to apply the same change there as well (i.e. external compilation, mps classloader and no extensions). It seems that those settings were a bit different almost in every case.

PR #1170, opened 2025-03-03
querylist: inline editor expressions improvements

wsafonov:
`QueryListNodeExpression` provides access to the input node of the query, but due to its ambiguous alias (`node`) conflicts with the `ConceptFunctionParameter_node` parameter of the query function, which makes it easy to pick up the wrong expression without directly noticing this. Also the code using both expressions at the same time becomes confusing.

Furthermore, `QueryListNodeExpression` and `QueryListIndexExpression` are currently allowed to be used outside of the querylist inline editor component, i.e. when there is no loop through the query list yet and no need for accessing index and querylist input node. Also, using the expressions in the nested querylist resulted in a wrong node and index to be returned: if used in the nested querylist query, the values were incorrectly returned for the nested querylist, but not for its owning parent querylist.

This PR provides the following changes to address the above issues:

- use unique, consolidated aliases for both expressions: `queryListInput` and `queryListIndex` 
- display short information about these available expressions in the querylist editor inspector to hint the user at this feature
- restrict usage of both expressions only inside of the querylist inline editor
- process the location of the expression (in the querylist editor or other parts of the querylist) and return the values from the correct querylist context

### DIFF
--- a/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
+++ b/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
@@ -23628,6 +23628,11 @@
             </node>
           </node>
         </node>
+        <node concept="1SiIV0" id="4EGFz66TUtk" role="3bR37C">
+          <node concept="3bR9La" id="4EGFz66TUtl" role="1SiIV1">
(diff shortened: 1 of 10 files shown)


alexanderpann (changes requested):
Two small things:

1.IconCell#relayoutImpl looks wrong.
2. You are targeting the master branch (= against MPS master) instead of the 2024.1 branch.

I really like the inline documentation. I also thought about something similar before.

alexanderpann (approved):

PR #1251, opened 2025-03-26
Add documentation on how to set up solution/model with merge language

klemensschindler:
Some extra setup is needed to make the merge language work in your solution.

Thanks to @alexanderpann and @eugenschindler for explaining how to make things work.


alexanderpann (approved):
Thanks for the contribution!

PR #1268, opened 2025-04-03
avoid CCE when not AbstractModule

artem-tikhomirov:
Not all SModule instances are AbstractModule.
And to remove some legacy API methods, we'd need their overrides not to demand their presence.


sergej-koscejev:
Please make such changes to the latest maintenance branch, if possible (`maintenance/mps20243` in this case) and propagate them to master through the cascading merge mechanism.

This will help ensure that the code can be built with two MPS releases with as few changes as possible, which in turn helps reduce the effort to migrate to the newer MPS version in the future.

PR #1325, opened 2025-04-22
Adapt ModelMerge to support merging of variability models

mfsoliveira:
During the implementation of a merge policy for iets3.core variability it was identified the need to some changes in the model merger:
- partial model merging policy is required to prototype policies for large languages
- reduction of some errors to warnings
- add policy for root nodes

Moreover, some bugs were found:
- merging with enums
- merging of root nodes
- Some corner cases with references 
- etc.


dbinkele:
@mfsoliveira minor code improvement: Let the method return the Object and use it for explicitly setting the variable on the level of calling the method, instead of doing this inside the method: 
image

dbinkele:
@mfsoliveira This is very identical code: 
image

Only the 1st code line is different?!

image

You probably can replace the 2nd line by the 1st one and then swap the arguments... same outcome to be expected...

dbinkele:
You need the parameters  only for the first 2 lines to find out which roots exists in both sides. Move this code out to the place where you call the method. It is exactly the place where you handle the roots which exist only on the left or right. A first step partitions the roots in 3 categories: Left, Right, Both. Then the we run the according algorithmic methods on them. This makes the intention of the method clearer.
image

dbinkele:
Changes not related to 'Modelmerger2' must be reverted: 
image

dbinkele:
@mfsoliveira Question: In 'ModelMergingRunner.run(...)' the code makes a distinction between when 'root policy' is present and when it is not. On the other hand in 'ModelMerger.merge(...)' we make sure that there is at least a 'default root'. My question is: Why do we need to handle the 'root policy not present'-case in 'ModelMergingRunner' if 'ModelMerger' is already handling it? Q2: Why not moving the setting of a 'default root policy' into the constructor?

dbinkele:
@mfsoliveira I found the test: http://127.0.0.1:63320/node?ref=r%3Aa0e8eec9-3c43-4072-a029-dd3c52aaa564%28de.itemis.model.merge.simple.demo.test%40tests%29%2F5638016647223604648

http://127.0.0.1:63320/node?ref=r%3Aa0e8eec9-3c43-4072-a029-dd3c52aaa564%28de.itemis.model.merge.simple.demo.test%40tests%29%2F1009498194811503017

http://127.0.0.1:63320/node?ref=r%3Aa0e8eec9-3c43-4072-a029-dd3c52aaa564%28de.itemis.model.merge.simple.demo.test%40tests%29%2F545364740423002515

In all 3 of them you use the 'TestingMerger', with which you manipulate given input data from the test and then the test is run. This makes the tests very hard to verify. Ideally the tests are self-containing with a simple Given-When-Then-Structure without any data massaging in between. I admit that sometimes it is hard to place all data inside the test, but then simple pointer to data outside should be preferred as it keeps the tests simple...........

mfsoliveira:
> @mfsoliveira it seems this newly added root is not working; http://127.0.0.1:63320/node?ref=r%3A1b4d0192-d370-48f1-b263-5be8f4cde603%28de.itemis.model.merge.simple.demo.merge.exec%29%2F8932628167883640190
> 
> Shouldn't there be some tests?

We do not need to test the ModelMergeConfiguration with pointers to models and policy name. I passing models and policy direct into test method. The policy is covered by test case MergingCollectionWithReference

mfsoliveira:
> @mfsoliveira minor code improvement: Let the method return the Object and use it for explicitly setting the variable on the level of calling the method, instead of doing this inside the method: image

I would prefer to remove that "if" completely. But do not want to define defaults for any marge policy, because it may cause failure silently. Therefore I decide to maintain some backward compatibility. If such compatibility is not needed, we just need to figure out the defaults for the tests and add it. (As you did - should we remove the "if" now?)

mfsoliveira:
> @mfsoliveira This is very identical code: image
> 
> Only the 1st code line is different?!
> 
> image
> 
> You probably can replace the 2nd line by the 1st one and then swap the arguments... same outcome to be expected...

There is margin for improvement, but I like to have locality of the data a method uses. Extract 2 lines of code to add 2 lines of code to call it far from the model, is not correct in my point of view. It does not removed duplication either, thus I let the preparation of the data required for the method in the method. But some of your changes improved it. We will find a trade-off in the next changes.

mfsoliveira:
> stinction between when '

Thanks
You fixed it. It was garbage from trying to make changes without braking the tests. Now with the root policy for all tests and default policy in ModelMerger we do not need it any more.

mfsoliveira:
> @mfsoliveira I found the test: http://127.0.0.1:63320/node?ref=r%3Aa0e8eec9-3c43-4072-a029-dd3c52aaa564%28de.itemis.model.merge.simple.demo.test%40tests%29%2F5638016647223604648
> 
> http://127.0.0.1:63320/node?ref=r%3Aa0e8eec9-3c43-4072-a029-dd3c52aaa564%28de.itemis.model.merge.simple.demo.test%40tests%29%2F1009498194811503017
> 
> http://127.0.0.1:63320/node?ref=r%3Aa0e8eec9-3c43-4072-a029-dd3c52aaa564%28de.itemis.model.merge.simple.demo.test%40tests%29%2F545364740423002515
> 
> In all 3 of them you use the 'TestingMerger', with which you manipulate given input data from the test and then the test is run. This makes the tests very hard to verify. Ideally the tests are self-containing with a simple Given-When-Then-Structure without any data massaging in between. I admit that sometimes it is hard to place all data inside the test, but then simple pointer to data outside should be preferred as it keeps the tests simple...........

The way I use it is good to keep all test data into test method. I do not manipulate the nodes I just put them dynamically into temporarily models. It is easy to see the data used in the test, than out side models.

dbinkele (approved):

PR #1328, opened 2025-04-28
Backporting batik.stubs and xml.stubs

arimer:
To fix https://github.com/mbeddr/mbeddr.core/issues/2844 we want to backport all relevant changes from regarding batik from `maintenance/mps20241` to `maintenance/mps20232`.


build.gradle
```diff
@@ -271,6 +275,39 @@ afterEvaluate {
 ext["itemis.mps.gradle.ant.defaultScriptArgs"] = defaultScriptArgs
 ext["itemis.mps.gradle.ant.defaultScriptClasspath"] = buildScriptClasspath
 
+task resolved_diagram_dependencies(type: Copy) {
```
sergej-koscejev, line 278: I think we should keep the createResolveTask() invocations below and remove these methods.

build.gradle
```diff
@@ -410,7 +462,7 @@ task cleanMps(type: Delete) {
     delete fileTree(dir: projectDir, include: ["**/classes_gen/**", "**/source_gen/**", "**/source_gen.caches/**", "tmp/**", "artifacts/**"])
 }
 
-cyclonedxBom { 
+cyclonedxBom {
```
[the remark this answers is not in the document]
sergej-koscejev: You added xml_libs in this PR so you should add it to the BOM as well?

sergej-koscejev (changes requested):

build.gradle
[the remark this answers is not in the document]
sergej-koscejev: I see what the question is about now. Yes, I think it should have been added in the original commit as well. (If there was a reason not to add it, then there should be a comment about that.)

build.gradle
arimer: I was thinking the same, but was not sure whey they were changed in the first place, so I kept them...

arimer: `resolved_diagram_dependencies(type: Copy)` was of type copy, is it save to be changed to `Sync`?

arimer:
@sergej-koscejev I pushed the required changes. Pls have a second look

sergej-koscejev (approved):

arimer:
Edit: It was just not yet published..

PR #1331, opened 2025-04-29
tables: extract gridmodel tests to a separate solution

sergej-koscejev:
This eliminates indirect dependency of tables.runtime on testing infrastructure.

### DIFF
--- a/code/.mps/modules.xml
+++ b/code/.mps/modules.xml
@@ -216,6 +216,7 @@
       <modulePath path="$PROJECT_DIR$/tables/languages/de.slisson.mps.tables/de.slisson.mps.tables.mpl" folder="tables" />
       <modulePath path="$PROJECT_DIR$/tables/languages/de.slisson.mps.tables/runtime/de.slisson.mps.tables.runtime.msd" folder="tables" />
       <modulePath path="$PROJECT_DIR$/tables/languages/de.slisson.mps.tables/sandbox/de.slisson.mps.tables.sandbox.msd" folder="tables" />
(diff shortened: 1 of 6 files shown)


arimer (changes requested):
It looks like you removed a `run tests from module` entry and did not add the newly created test solution [here](http://127.0.0.1:63320/node?ref=r%3Afb24ac52-5985-4947-bba9-25be6fd32c1a%28de.itemis.mps.extensions.build%29%2F7547442689024766459): 

grafik

could you pls. double check?

arimer:
Also, why dont we want this in 2023.2?

sergej-koscejev:
> It looks like you removed a `run tests from module` entry and did not add the newly created test solution [here](http://127.0.0.1:63320/node?ref=r%3Afb24ac52-5985-4947-bba9-25be6fd32c1a%28de.itemis.mps.extensions.build%29%2F7547442689024766459):

Both are part of `de.slisson.mps.all.tests` group which is included.

sergej-koscejev:
> Also, why dont we want this in 2023.2?

The context of this is that 2024.1 has more strict behavior at run time than 2023.2 and may refuse to load a plugin if the runtime of a used language is missing even if it might not be actually used. @artem-tikhomirov is implementing additional checks for this in 2024.1 in the build language ([MPS-38286](https://youtrack.jetbrains.com/issue/MPS-38286/Build-language-does-not-detect-plugin-dependency-from-used-languages)). Once these checks are in place, the tables plugin would have to declare a dependency on unit testing infrastructure unless we remove it.

But in 2023.2 this is not needed and I don't want to touch older versions unnecessarily.

arimer (approved):

PR #1340, opened 2025-05-06
richtext: fixed RichTextUtil for Text creation

wsafonov:
Fixed utility class for creating instances of `Text` to avoid space ` ` before the actual string content. This might be helpful when requiring exact `Text` content, for example in nodes compare scenarios.

_Note_: fixed method  `RichTextUtil.createTextFromSingleString()` is only available as utility for the external usage, but not used in the richtext language itself. Therefore the change has no impact on the language behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project are documented in this file.
 
 The format is *loosely* based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) .The project does *not* follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## May 2025
+
+### Fixed
+
+- *de.slisson.mps.richtext* RichTextUtil.createTextFromSingleString() creates `Text` instance without unexpected prepended space.
+
 ## April 2025
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


arimer:
Why not applieying the fix to 2023.2?

wsafonov:
> Why not applieying the fix to 2023.2?

The impact is very small, the utility has been there for ages and it didn't care anyone - I don't expect some urgently requiring it for older versions. So I thought why not saving some efforts and doing this only for the relevant branch.

arimer (changes requested):
- You removed the `textNode.ensureNormalized()` call, why?
- Please update the change log

wsafonov:
> * You removed the `textNode.ensureNormalized()` call, why?

Normalization is not needed if there is only one word (which is what we want in this simple case).

> * Please update the change log

Done

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project are documented in this file.
 
 The format is *loosely* based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) .The project does *not* follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## May 2025
```
arimer, line 7: I guess you did not re-generate this file? 
To do it as intended you need to update some model content and re-run the gradle task as mentioned in the readme: https://github.com/JetBrains/MPS-extensions?tab=readme-ov-file#creating-a-pr

Otherwise your change log update will be overriden next time...

arimer (changes requested):

arimer (approved):
Fly me to the 🌔

Issue #1339, opened 2025-05-06
Add generator test language


enikao:
There is no good way to test generators in MPS. The [built-in generator test](https://www.jetbrains.com/help/mps/testing-languages.html#generatortests) lack in several aspects:

* Output is text-only, so hard to read
* No way to post-process the generated output before comparison, or ignore some parts
* Ordering of nodes is an issue
* No meaningful names in test run view
* No way to compare text output
* Implementation is hard to extend

Thus, we should have a proper language to test generators.
Thankfully, `com.mbeddr.mpsutil.generatorfacade` and `com.mbeddr.mpsutil.compare` already provide a good base.

# Requirements
DSL that makes common things simple and complex things possible. Reading a typical test should take less than 5 seconds.
It should run in both MPS interactively, and headless (build pipeline).
If run interactively, diffs should be shown in a diff view.

## Model comparison
* Compare the generation result of 
  * a single (root) node with a specified reference node
  * a whole model with a specified reference model
* With genplan
  * default (as defined by the input model)
  * explicitly specified
* Optionally ignore
  * configured properties
  * configured references
  * features of `BaseConcept`
  * Specific concept instances (e.g. `TestInfo`)
* Optionally adjust node ordering
  * as-is
  * by some property (default: `INamedConcept.name`)
  * by some custom criteria
* Optionally allow custom code to post-process the generated model before comparison (e.g. to replace timestamps or random ids)

## Text comparison
* Compare the textual generation result of 
  * a single (root) node with a specified reference text
  * a whole model with a specified reference directory potentially containing several files and sub-directories
* With genplan
  * default (as defined by the input model)
  * explicitly specified
* Easy way to point to the reference directory that works both interactive locally and headless in a build pipeline
* Optionally normalize line endings
* Show mismatching file contents
* Show missing files or directories on either side (reference directory vs. generation result)
* Optionally allow custom code to post-process the generated text before comparison (e.g. to replace timestamps or random ids)

related: #490, #188

sergej-koscejev:
@alexanderpann @enikao FYI the current generator of the compare language uses RepositoryHelper from its own behavior in the generated code. I think you don't need any helpers in tests but can just use `project.getRepository()`  where `project` is a `jetbrains.mps.lang.test.structure.ProjectExpression`. If that does not work for some reason, then RepositoryHelper should be moved to a separate solution and the solution should be specified as runtime of the language.

[...]

PR #1343, opened 2025-05-07
Feature: Generator Tests

alexanderpann:
See https://github.com/JetBrains/MPS-extensions/issues/1339 for more discussion why the MPS generator tests are insufficient. The language *com.mbeddr.mpsutil.compare* has a new assert statement *assert generated model* to generate models or single nodes and compare them in tests. TextGen is also supported. Referenced nodes must be instances of richtext Words when the latter option is used. More option can be set in the inspector.

Scherm­afbeelding 2025-05-07 om 13 13 09

Scherm­afbeelding 2025-05-07 om 13 16 40


enikao:
For me, it would make more sense to separate the `assert generated model` from `assert generated text model`. It's very rare that I want to change the one to the other, and the text variant is now hard to discover (doesn't show up in autocomplete for `assert` statements).

enikao:
Type error is shown on the wrong child:
image

enikao:
Typecheck validates for `sequence<node<>>` but generator doesn't like it:

```
test generatorModelInsteadOfModelPointer { 
  sequence<node<>> expectedNodes = new node<>[]{expectedClass, expectedClass2}.asSequence; 
  assert generated model model-ptr/com.mbeddr.mpsutil.comparator.diff.demo.tests.input/.resolve(model/.getRepository()) equals expectedNodes with plan CustomPlan; 
}
```

```
Compilation problems
incompatible types: java.lang.Iterable<org.jetbrains.mps.openapi.model.SNode> cannot be converted to java.util.List<org.jetbrains.mps.openapi.model.SNode> (MPS-extensions\code\compare\solutions\com.mbeddr.mpsutil.comparator.diff.demo.tests\test_gen\com\mbeddr\demo\comparator\diff\tests\diffs\GeneratorTest_Test.java:381)
```

enikao:
In subsequent PRs, we might want to systematically look into replicating the _match options_ of MPS generator tests.

enikao:
I don't think the tests in `GeneratorTest` actually test whether the additional parameters (genplan, stuff in inspector) are taken into account.

enikao:
Would it be feasible to have an `assert <node> generated text equals <word>` statement? Symmetric to `AssertNodeEquals`, but for text?

enikao:
Would it be feasible to show the diff editors of `AssertNodeEquals` as tabs, not new windows?

enikao:
Why no diff viewer for `AssertGeneratedModelEquals`? Ideally, we would see _all_ contents of the model in one diff viewer.

alexanderpann:
> I don't think the tests in GeneratorTest actually test whether the additional parameters (genplan, stuff in inspector) are taken into account.

No idea how I would even test them. For me, those tests are rather smoke tests to check that the generator doesn't work. Most of the time we only have those kind of tests here.

alexanderpann:
> I assume AssertNodeEquals doesn't try to show diff viewers in headless mode, even if the checkbox is activated?
No, it doesn't. When there is no UI, only the text diff is shown.

alexanderpann:
> Would it be feasible to have an assert <node> generated text equals <word> statement? Symmetric to AssertNodeEquals, but for text?

This sounds like just an alias for assert generated model with this specific use case, right? Or do you just want to call textgen without the generator?

alexanderpann:
> Would it be feasible to show the diff editors of AssertNodeEquals as tabs, not new windows?

You can create a separate ticket for that, but it would not be easy. The class StructDifferenceDialog is used to show a dialog. We would have to check how the commit dialog is implemented and replicate those Commit tabs in the editor. They also support applying the changes so this would cover the last feature request.

enikao:
> > Would it be feasible to have an assert  generated text equals  statement? Symmetric to AssertNodeEquals, but for text?
> 
> This sounds like just an alias for assert generated model with this specific use case, right? Or do you just want to call textgen without the generator?

I see two axis there:

```
              node compare                    text compare
single node   AssertNodeEquals                ?
model         AssertGeneratedModelEquals      AssertGeneratedTextModelEquals
```

enikao:
> > I don't think the tests in GeneratorTest actually test whether the additional parameters (genplan, stuff in inspector) are taken into account.
> 
> No idea how I would even test them. For me, those tests are rather smoke tests to check that the generator doesn't break. Most of the time, we only have this kind of tests here.

I added some tests to show how I think it can be done. This is sufficiently complex logic, so I think they warrant tests.

The three tests I wrote immediately uncovered some usability issues:

* Writing _ignored properties_ is rather tidious:
  image

* `model` might not be the best parameter name:
  image

alexanderpann:
> I see two axis there:

I don't get it. You always use the same statements for comparing nodes and models, for AssertNodeEquals you have to iterate through a loop, the other ones support different expected parameters.
Node comparison: AssertNodeEquals
Generated model: AssertGeneratedModelEquals
TextGen: AssertGeneratedTextModelEquals

enikao (dismissed):

PR #1362, opened 2025-05-21
mpsutil.compare: inline repository calls

alexanderpann:
See https://github.com/JetBrains/MPS-extensions/issues/1339#issuecomment-2898370042. Using the `project` expression was actually my first attempt, but it didn't work because of the generator priorities.


sergej-koscejev:
Found the reason for 'generates into' lang.generator. There's an invisible BaseCommentAttribute with some if macro or something like that in reduceAssertGeneratedModel on [this node](http://127.0.0.1:63320/node?ref=r%3A4e2e8730-5cdd-4a73-9b92-3f36ecd3ae72%28com.mbeddr.mpsutil.compare.generator.template.main%40generator%29%2F5440956104040849703), visible in the node explorer.

sergej-koscejev:
I understand the priority problem now, lang.test is a top-priority group whereas compare is not. And also `assert node equals` may be used outside of a node/editor test where ProjectExpression (or the code it generates into) is not available.

I think a refactoring is in order. Project should be passed explicitly to `assert node equals`, except when it is used inside a node/editor test case. In that case we should be able to use baseLanguageInternal to simply generate `myProject` to get the current project, this way we won't need to depend on the lang.test generator.

Project should also be passed to `DiffView#showDiffDialog` so that it doesn't have to use the ugly `ProjectManager.getInstance().getOpenedProjects().get(0)` hack.

Another issue I see is that there is a generator IF macro conditioned on `node.isDiffEnabled` and that behavior method evaluates `isHeadlessEnvironment()` so it only makes sense to call it at run time, not at generation time.

sergej-koscejev (approved):
Approving the PR now in order to unblock progress.

PR #1400, opened 2025-06-01
compare: fixed duplicated code generation for assertNodeEquals

wsafonov:
Using COPY_SRC inline for expected/actual nodes could lead to duplicated code, e.g. in case of quotations as argument values, since expected/actual node arguments are then reduced twice (for compare and for diff calls). 

To avoid this, expected/actual node arguments have been extracted into local variables


arimer:
To give some more background info. This change fixes iets3.os errors like those:

https://mps.builds.itemis.cloud/buildConfiguration/Iets3_BuildOrgIets3core/380645?showLog=380645_6858_438.6609.6858.6919.7111&logFilter=debug&logView=flowAware

arimer (approved):
Thanks for the fixes 🌹

wsafonov:
@alexanderpann let me know if you have any comments post-merge - we had to act quickly to provide this fix for the iets3.os builds.

PR #1420, opened 2025-06-04
renovate: group all third-party Gradle dependencies

sergej-koscejev:


sergej-koscejev:
Merging without review and build because of non-code changes

PR #1439, opened 2025-06-05
Refactor uses of ReloadableModule

artem-tikhomirov:
Address uses of classloading through ReloadableModule. Now, we keep classloading aspect of a module independent of SModule class hierarchy (and independent of any repository read/write access, too).

Uses of ReloadableModule.getOwnClass() have been refactored with ModuleRuntime.getOwnClass(). ClassLoaderManager listeners (low-level, module/repository-kind of notification),  have been replaced with ModuleDeploymentListener of LanguageRegistry.


sergej-koscejev:
I don't think we want to be defensive about loaded classes not being assignable to expected classes but rather fail with an error or an exception in this case.

E.g.

```
Class<?> descriptorClass = module.getOwnClass(className); 
if (descriptorClass == null || !ISelectionIntentionsDescriptor.class.isAssignableFrom(descriptorClass)) { return; }
...
```

should be just

```
Class<?> descriptorClass = module.getOwnClass(className); 
if (descriptorClass == null) { return; }
```

sergej-koscejev:
@artem-tikhomirov I'm going to force-update this PR myself.

PR #1454, opened 2025-06-12
mpsutil.intentions: Fix an issue with intentions multiplying in the intentions menu

alexanderpann:
Do not populate the intention actions with a default action. It leads to intentions multiplying when the default intention action provider is deregistered through the extension point `IntentionActionsProvider.EP_Name`.

The issue and the fix can be testing with a new project plugin with the init block:
`IntentionActionsProvider.EP_NAME.getPoint().unregisterExtension(IntentionActionsProviderImpl.class);`


sergej-koscejev (changes requested):
The code is hard for me to follow and I don't understand why the fix actually fixes the problem.

Perhaps the handling of actions could be put into a separate class and the methods could be ordered roughly according to the call tree (if method A calls B which calls C then they should ideally be listed in the same order).

There are two methods called getIntentionsGroup which seem to be unrelated. They should be given different names.

From some short debugging it seems that the problem is related to caching because the duplicate intentions accumulate in the cache and are added to the existing ones.

This code:
```
foreach entry in getCache().asMap() { 
  string groupName = entry.key; 
  groups.putAll(groupName, entry.value); 
}
```
could be simplified to `groups.putAll(getCache())`, right?

In `processActionGroup`, why are actions both added directly to `groups` and also taken from the cache? What is the purpose of the cache?

sergej-koscejev:
Why is substituteAction necessary? It seems to be replacing an action with its clone, but why? It also seems to be putting the old action into the cache, why?

PR #1462, opened 2025-06-17
Rewrite custom intentions

sergej-koscejev:
* Simplify the implementation of custom intentions
* Fix a typo
* Use the term 'section' rather than 'group' in the context of intention menu, to avoid confusion with action groups.


alexanderpann (changes requested):
In `addGroupAnnotationToIntention` the description has to be adjusted. Other than that the PR looks good. I tested it with the sandboxes and by unregistering IntentionActionsProviderImpl.

alexanderpann (approved):

Issue #1488, opened 2025-07-01
assert equals statement ignores node attributes


enikao:
I compare two nodes with _left_ having one node attribute and _right_ having two attributes.

`assert equals` considers them equal, even though they are not.

Issue #1456, opened 2025-06-13
assert equals statement sometimes fails with missing model for node when detecting difference and "show diff" is enabled


enikao:
We're using `com.mbeddr.mpsutil.compare.structure.AssertNodeEquals` with the _actual_ node being just created, not part of a model yet. If the assertion fails and we enable "show diff", we don't see a diff, but an exception.

(stacktrace to be added later from different system)

PR #1500, opened 2025-07-10
Improvements to comparator and compare language

alexanderpann:
Fixes
- #1456 
- #1488


kbirken:
@alexanderpann @enikao This PR changed the behaviour of "assert equals" wrt. node attributes. If tests do not care for differences of node attributes, they will break now. Unfortunately, this breaks downstream projects - and there seems to be no configuration option to fix it (e.g., a way to ignore node attributes similar to ignoring properties or references). What can we do to avoid turmoil in downstream projects?

PR #1519, opened 2025-07-28
build: use com.specificlanguages.mps plugin

sergej-koscejev:
The plugin provides a DSL and some quality-of-life improvements such as reusing MPS versions with Gradle artifact transforms, specifying the JBR version as a proper Gradle dependency, generating build scripts without the need for an 'all scripts' build script, etc.

The current plugin version is 2.0.0-pre4 and I am going to release a final version soon after I update the docs. But there is still opportunity to make some breaking changes to the build DSL if necessary.


sergej-koscejev:
FYI @nkoester

wsafonov (changes requested):
Thanks for this quite a legendary refactoring! The new DSL looks clean and self-explaining. I would say it is the right way to go. As far as I could review the concepts from the new plugin, everything seemed solid to me in order to make a swich to the new DSL. So let's give it a try. Maybe in fact there is even no need for further breaking changes.

I just saw we could reuse repeated path specification in many places - see inline comment.

Issue #1531, opened 2025-08-05
Add efficient NodeHasher implementation


enikao:
We developed an efficient NodeHasher implementation.
It would be useful as a generic tool.
Depends only on Google Guava Murur3 and SNodes et al. 

```java
/** 
  Hashes all passed nodes by their concept, property+value, (deep) containments+node, and references+target. 
   
  Can ignore {@link setIncludeNodeIds() node ids}, {@link setIncludeBaseConceptProperties() BaseConcept properties}, {@link setIncludeAnnotations() annotations/node attributes}, and {@link ignore() arbitrary features}. 
  By default, everything is included. 
   
  If we consider node ids, we compare reference targets by node id; otherwise by the target node's {@link jetbrains.mps.smodel.SNode#getPresentation() }. 
 */ 
 
public class NodeHasher { 
   
  private static final string UNSET_PROPERTY_VALUE = "false"; 
   
  private final nlist<> nodes; 
  private final Murmur3_32Hash.Murmur3_32Hasher hasher; 
  private final set<SProperty> ignoredProperties = new linked_hashset<SProperty>; 
  private final set<SContainmentLink> ignoredContainments = new linked_hashset<SContainmentLink>; 
  private final set<SReferenceLink> ignoredReferences = new linked_hashset<SReferenceLink>; 
   
  private boolean includeNodeIds = true; 
  private boolean normalizeBooleanProperties = true; 
   
  /** 
    Hashes all nodes. 
     
    @param nodes Nodes to hash. The order of nodes, and (in)direct children, matters.                                                     
   */ 
   
  public NodeHasher(nlist<> nodes, int seed)  { 
    this.nodes = nodes; 
    hasher = new  Murmur3_32Hash.Murmur3_32Hasher(seed); 
  } 
   
  /** 
    Convenience factory to hash nodes with default seed, including node ids and all features. 
     
    @param nodes Nodes to hash. The order of nodes, and (in)direct children, matters. 
   */ 
   
  public static int hash(node<> ... nodes)  { 
    new  NodeHasher(nodes.asSequence.toList, 0x21375341).hash(); 
  } 
   
  /** 
    Whether we consider <tt>BaseConcept.shortDescription</tt> and <tt>BaseConcept.virtualPackage</tt> properties while hashing. 
   */ 
  public void setIncludeBaseConceptProperties(boolean includeBaseConceptProperties)  { 
    if (includeBaseConceptProperties) { 
[...]

PR #1532, opened 2025-08-05
added efficient NodeHasher

enikao:
implements #1531 

I put it into the `compare` plugin, as comparing and hashing nodes are quite similar (cf. Java `Object.equals()` and `Object.hashCode()`). This introduces a new dependency to `de.q60.mps.collections.libs`.

I used namespace `nl.f1re.mpsutil.hasher`.

Both of this can be adjusted.

### DIFF
--- a/code/.mps/.gitignore
+++ b/code/.mps/.gitignore
@@ -1,2 +1,5 @@
 workspace.xml
 product-workspace.xml
+
+inspectionProfiles/
+shelf/
--- a/code/.mps/modules.xml
+++ b/code/.mps/modules.xml
@@ -75,6 +75,8 @@
       <modulePath path="$PROJECT_DIR$/hacks/solutions/de.itemis.mps.nativelibs.loader/de.itemis.mps.nativelibs.loader.msd" folder="hacks" />
       <modulePath path="$PROJECT_DIR$/hacks/solutions/de.slisson.mps.hacks.editor/de.slisson.mps.hacks.editor.msd" folder="hacks" />
       <modulePath path="$PROJECT_DIR$/hacks/solutions/de.slisson.mps.reflection.runtime/de.slisson.mps.reflection.runtime.msd" folder="hacks" />
+      <modulePath path="$PROJECT_DIR$/hasher/solutions/nl.f1re.mpsutil.hasher.tests/nl.f1re.mpsutil.hasher.tests.msd" folder="hasher" />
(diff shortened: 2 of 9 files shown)


enikao:
The list of dirty files that failed the build look very similar to the changes from "force save all": https://github.com/JetBrains/MPS-extensions/pull/1532/commits/9807359800740b54e64bc45a249ed5d5388c0fca

Is there some process issue?

alexanderpann:
It should work fine. My guess is that you didn't have the model merger languages build when you executed force save all and you got some incorrect changes as a result. As I mentioned before: Just apply the diff from the build log via the method described in the readme

sergej-koscejev:
Also, why specifically murmur3 and why 0x21375341 as the default seed?

enikao:
> The implementation handles missing or invalid concepts defensively, IDs are used where available, falling back to names. This means that the hash of a node will change when its language is available vs unavailable. For references, hashing may use the presentation of the target node, which introduces even more dynamic dependencies (not only the target language has to be available but also whatever dependencies the behavior method has).

The alternative code path where meta-elements are not an instance of `...Adapter` are very theoretical -- I wouldn't know how to trigger that code. So we could throw an exception.

> 
> Is this intentional? I can imagine that in some cases "aggressive" handling would be preferred, i.e. always use IDs and throw an exception if some ID is not available. In the worst case, if both approaches are useful, this could be made configurable.
> 

The user can chose whether we include node ids (for both regular nodes and reference targets) with `setIncludeNodeIds()`.
The fallback to target presentation / resolveInfo is IMHO the best we can do.

We could change visibility of `hashNode()` / `hashProperty()` / `hashReference()` to `protected`, then a user could create a specialized subclass for their uncovered use-case.


> What is the use case that motivated you to add this hasher?

Checking whether there's any change in a subtree.

enikao:
I'm close to giving up on this. I have a very hard time managing all of chasing `main`, satisfying TeamCity, conjuring git, migrating in MPS, remigrating in Gradle. After I found my way through all of that, there's a new change on `main`, and I can restart.

sergej-koscejev:
@enikao I have cleaned up the PR, moving the changes in .gitignore to `code/.mps/.gitignore` where I think they belong.

enikao:
> @enikao I have cleaned up the PR, moving the changes in .gitignore to `code/.mps/.gitignore` where I think they belong.

Thanks a lot!

alexanderpann (approved):

PR #1549, opened 2025-08-19
nl.f1re.mps.editor.swing: New language for customizing swing components

alexanderpann:
This language add a new swing component cell that can be customized and also sets some sensible default values for the component:

- **component provider**: created the swing component
- **style function**: this function influencs the font size and the general style of the component
- **string representation**: customize the plain text representation of the component when pasting to the clipboard
- **overwrite default ascent**: Overwrite the suboptimal default MPS ascent calculation which is used to horizontally align the component in the editor. Set this flag to false when the swing component doesn't have `getAscent` properly implemented.

Bildschirmfoto 2025-08-19 um 10 39 45

Bildschirmfoto 2025-08-19 um 10 41 36

- This PR is a follow-up of 
	- https://github.com/JetBrains/MPS-extensions/pull/1314
	- https://github.com/mbeddr/mbeddr.core/pull/3087
- It implements/fixes
	- https://youtrack.jetbrains.com/issue/MPS-38915/Rendered-text-customizations-for-swing-components
	- https://youtrack.jetbrains.com/issue/MPS-36613/EditorCellComponent-inconsistent-look-with-rest-of-editor
	- https://youtrack.jetbrains.com/issue/MPS-36612/CellModelJComponent-Missing-common-properties

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ The format is *loosely* based on [Keep a Changelog](https://keepachangelog.com/e
 
 ### Changed
 
-- *nl.f1re.mps.editor.swing* The experimental stylesheet `IntelliJStyleSheet` was moved to this language. It emulates the style of the IntelliJ UI.
+- *nl.f1re.mps.editor.intellij* The experimental stylesheet `IntelliJStyleSheet` was moved to this language. It emulates the style of the IntelliJ UI.
 
 ## July 2025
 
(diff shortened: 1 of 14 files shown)


sergej-koscejev (changes requested):
Please exclude unrelated changes from this PR (mkdocs changes, changes in the styles language).

Are IntelliJStyleSheet and FontHelper somehow related to Swing components? Why are they in the Swing language? Where do the magic numbers for font styles come from? This should be documented in that class. Also, shouldn't the offsets be in the form of multipliers rather than additions?

Same for MPSOnOffButton - in this case I can see why it's in the Swing language but I would like to see some comment in the code about the baseline computation, what's wrong with the default one?

Regarding TextHelper, in case for JTabbedPane is it really the best idea to use tab titles separated by commas? Do we really need a case for this component? What is it useful for?

Please add a doc comment for EditorCell_IntelliJComponent describing the purpose of the class (why is EditorCell_Component not good enough?).

I don't like that the class is never used as-is but always extended with a toString() method. Perhaps it would be better if the toString code would be passed to the class as a closure, if specified.

`ComponentUtil.putClientProperty` is deprecated, please use a non-deprecated method as this is new code.

sergej-koscejev (approved):

PR #1566, opened 2025-09-01
ModelMerging: Organize the project 

mfsoliveira:
This PR try to organize the model merging package in the MPS-Extension before improve testing:

- delete partial deleted languages and solutions that has no longer the definition file but keep their model files
- rename solutions to be more expressive about what it is testing and use test.de.itemis.model.merge prefix like all other tests solutions.
- organize test languages and test solutions in virtual folders.
- rename modelmerger folder to explicit express it is deprecated (shall I delete it?)
- create a modelmerging folder and move its languages and solutions into there


sergej-koscejev:
@mfsoliveira I will delete the branch in two weeks unless you tell me not to.

PR #1579, opened 2025-09-07
polymorph func: use ModuleRuntime instead of deprecated ReloadableModule

sergej-koscejev:
* Replace ReloadableModule w/ ModuleRuntime in DescriptorCache.
* ImplementationsFromFunctionModule: use ModuleRuntime to access generated classes.

Based on a contribution by Artem Tikhomirov (#1439).


alexanderpann (approved):
LGTM. The AtomicReference in DescriptorCache looks unecessary but doesn't hurt. There is no access to this variable from multiple threads.

PR #1598, opened 2025-09-19
mps.tables: Support copying, pasting and deleting of table selections

alexanderpann:
Tables now support copying, cutting and deleting when multiple cells are selected with the mouse. Add the action map *TableSelectionActionMap* in the inspector of the table and implement the extension point *TableCopyPaste* to support these features for a specific table. In Addition, two new intention are available for tables that implement the extension point *DataTransformation* which allow parsing table data in textual form (comma- or tab-separated) and paste it to table (e.g., 10 as a number literal).

Note: This feature was initially implemented by @svott7. I moved it to MPS-Extensions and modified it a bit to be more generic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is *loosely* based on [Keep a Changelog](https://keepachangelog.com/e
 
 ## September 2025
 
+### Added
+
+- *de.slisson.mps.tables* Tables now support copying & pasting, cutting and deleting when multiple cells are selected with the mouse. Implement the extension point *TableCopyPaste* to support these features for a specific table.
+- *de.slisson.mps.tables* Two new intention are available for tables that implement the extension point *DataTransformation.* *It* allows to parse table data in textual form (comma- or tab-separated) and paste it into the table (e.g. 10 as a number literal).
+
 ### Fixed
 
 - *de.itemis.mps.spellcheck* An exception coming from the WordsToDictionaryIntention was fixed.
--- a/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
+++ b/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
@@ -4295,6 +4295,11 @@
             </node>
           </node>
         </node>
+        <node concept="1SiIV0" id="7NamNJWLm3U" role="3bR37C">
+          <node concept="3bR9La" id="7NamNJWLm3V" role="1SiIV1">
+            <ref role="3bR37D" to="ffeo:3MI1gu0QouH" resolve="jetbrains.mps.editor.runtime" />
+          </node>
+        </node>
       </node>
       <node concept="1E1JtD" id="29so9Vb$6Th" role="2G$12L">
         <property role="BnDLt" value="true" />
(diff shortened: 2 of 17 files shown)


sergej-koscejev (commented):
1. CopyPasteSupport#setTable should be called setTableNode.
2. In TableSelectionActionMap there are multiple calls to support.setTable(node) that probably can be removed (e.g. [this one](http://127.0.0.1:63320/node?ref=r%3A1e59a084-7ebe-4e95-89ab-c58a7e396583%28de.slisson.mps.tables.editor%29%2F1206675692275615536))
3. TableCopyPaste: should it rather have an `isApplicable(node)` method than `getConcept()`? What if it's also applicable to subconcepts or some nodes but not others?
4. I'm not quite sure about classloading issues with TableCopyStorage being an IDEA lightweight service. What's the benefit as opposed to having a static variable? The comment says the data will be preserved after MPS exit but I don't see how it's achieved.
5. The storage should probably be cleared on copy-paste (including from some other application) and possibly when the nodes are unloaded? And/or it probably should store SNodeReferences rather than SNodes?
6. TableData.iterateRowwiseWithGaps looks complicated but also unused?
7. Can we make copy-pasting a default behavior so that action maps are only required to override it with something custom?
8. Would it be possible to implement node-to-text transformation for copying based on editor cells' renderText() instead of node presentation? This combined with the previous point would make copy-paste work for many tables out of the box.

alexanderpann:
1. ✔️ 
2. ✔️ 
3. Also changed for DataTransformer
4. Changed. I had issues with the tests but apparently it is now also working this way.
5. I am clearing it now on every change of the clipboard contents. Not sure when else it should be cleared but I am storing nodes on purpose so that you could modify the nodes in the clipboard if needed (e.g. with a copy-paste handler in the future)
6. ✔️  Removed
7. The action map is now automatically set when no action map is used.
8. I am now only falling back to node representation when I can't find the node cell anymore.

sergej-koscejev (commented):
I don't like UndirectedTableRangeSelection extending from TableRangeSelection. I don't see the benefits of this. I'd rather go the other way and replace it with a TableRange class that's just a rectangle.

The change from row/column start/end to two intervals is problematic because the intervals are now mutable, making the selection mutable as well, whereas it was immutable before.

Instead of passing `SRepository` to methods like `CopyPasteSupport.paste` or `nodeDataToStringData` let's pass EditorContext. Also perhaps we should pass EditorContext to the new isApplicable methods? And in general to all methods that may be defined by the user, such as `CopyPasteSupport#copy` or `#delete`.

DataTransformer.isApplicate → isApplicable.

I was initially skeptical about the approach with the default action map. The pattern used so far is generating calls to EditorCell.setAction if there is no action map or the action map does not have a particular action defined ([example](http://127.0.0.1:63320/node?ref=r%3A00000000-0000-4000-0000-011c8959029f%28jetbrains.mps.lang.editor.generator.baseLanguage.template.main%40generator%29%2F4348485823408509922)). But I think in the end the default action map is a cleaner solution so let's keep it. The linked code does not take into account action map imports, for example.

Please also update the changelog to match the current state of the changes. And instead of 'In Addition' split the entry into two for clarity.

nkoester:
> * The transformation logic currently relies on the parser, which does not seem fully safe and likely needs proper escaping handling.

I just stumbled over this and why was this merged if the parser - which is the core of a copy/paste operation in this context - is not safe/fully working? the [included example over here](http://127.0.0.1:63320/node?ref=r%3A7b909ba9-e58e-4fbd-8829-ecfbd89930fa%28de.slisson.mps.tables.sandbox%29%2F1652594944444666763) simply does not work:

1. There can be multiple transitions in one table cell. This screws up [the counters used to decide what to copy](http://127.0.0.1:63320/node?ref=r%3Af1c5f6f4-f735-4c95-8571-0c9b99e0bba3%28de.slisson.mps.tables.demolang.plugin%29%2F1206675692272665246). As a result garbage is copied.
2. Partial copy/paste is broken: Copying empty cells (aka null) break the copy operation when copying without headers/somewhere in the middle. Same for partial pasting. This is due to the [transformer being null over here](http://127.0.0.1:63320/node?ref=r%3Ac386283f-4bfc-42ea-a1b4-65abe196bd30%28de.slisson.mps.tables.runtime.plugin%29%2F1206675692268408019).

Honestly, the tests of this feature do not test what is implemented here and miss both these errors. I am confused why this got merged even though the issues were know.

PR #1635, opened 2025-10-20
[MPS 2025.3] Fix javac compilation error: re-export dependency

fisakov:
A dependency on de.q60.mps.collections.libs solution is re-exported by de.q60.mps.polymorphicfunctions.runtime.
Also, explicitly list library jars in build layout.


fisakov:
Anyways, this patch makes mps-extension build on our CI. Just FYI.

sergej-koscejev:
> > I don't see why the change from 'files from' to individual files is useful, in my view it only adds maintenance burden in case the set of jars changes.
> 
> Disagree. Data in the model is searchable, reliance on external source of data is akin to blind man walk.

This data is derived, the origin of the data is in the Gradle build files (and/or transitive dependency POMs). Adding redundant data to the model that has to be manually updated just so that it becomes searchable is counterproductive.

fisakov:
Whatever, your house your rules.

PR #1639, opened 2025-10-23
[MPS 2025.3] Drop MPS.ThirdParty module completely

fisakov:
No dependencies in mps-extensions on this module.

### DIFF
--- a/code/testing/solutions/nl.f1re.testing.examples/models/nl.f1re.testing.examples@tests.mps
+++ b/code/testing/solutions/nl.f1re.testing.examples/models/nl.f1re.testing.examples@tests.mps
@@ -56,7 +56,6 @@
     <import index="6lvu" ref="1ed103c3-3aa6-49b7-9c21-6765ee11f224/java:jetbrains.mps.nodeEditor.cellMenu(MPS.Editor/)" />
     <import index="fnpx" ref="498d89d2-c2e9-11e2-ad49-6cf049e62fe5/java:com.intellij.notification(MPS.IDEA/)" />
     <import index="alof" ref="742f6602-5a2f-4313-aa6e-ae1cd4ffdc61/java:jetbrains.mps.ide.project(MPS.Platform/)" />
-    <import index="fnpy" ref="498d89d2-c2e9-11e2-ad49-6cf049e62fe5/java:com.intellij.notification(MPS.IDEA/)" />
     <import index="tp5g" ref="r:00000000-0000-4000-0000-011c89590388(jetbrains.mps.lang.test.structure)" />
     <import index="kz9k" ref="1ed103c3-3aa6-49b7-9c21-6765ee11f224/java:jetbrains.mps.openapi.navigation(MPS.Editor/)" />
     <import index="cj4x" ref="1ed103c3-3aa6-49b7-9c21-6765ee11f224/java:jetbrains.mps.openapi.editor(MPS.Editor/)" />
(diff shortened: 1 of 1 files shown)


alexanderpann:
The included jars in MPS.IDEA expose more libraries than `MPS.ThirdParty`. Unfortunately, the library is used in other platforms such as mbeddr (22 usages) and probably in customer projects, so this would be a breaking change without a migration. One of the main motivations was that some libraries use the Kotlin standard library which is not exposed as stubs that's why `MPS.Kotlin` was also added to MPS-Extensions which `MPS.ThirdParty` depends on (there is a separate issue that I figured out: https://github.com/JetBrains/MPS-extensions/issues/1640)
@fisakov You can open the class [CommitCheck](http://127.0.0.1:63320/node?ref=39983771-4e9b-401b-a1a9-1da6c777c843%2Fjava%3Acom.intellij.openapi.vcs.checkin%28MPS.ThirdParty%2F%29%2F~CommitCheck) where the reference to Continuation is broken. All classes from the IntelliJ with references to the standard library can't be used as a result. If this is fixed on MPS master, we can think of merging it but preferable with a migration.

fisakov:
>Unfortunately, the library is used in other platforms such as mbeddr (22 usages) and probably in customer projects, so this would be a breaking change without a migration.

I manually updated references in Mbeddr in https://github.com/mbeddr/mbeddr.core/pull/3219. Both mps-extensions and Mbeddr are passing our CI build. I can think of providing a migration, just not an automated migration. It will probably be a shell script with sed/awk commands. 

> One of the main motivations was that some libraries use the Kotlin standard library which is not exposed as stubs that's why `MPS.Kotlin` was also added to MPS-Extensions which `MPS.ThirdParty

The proper way to access Kotlin stubs is via Kotlin stubs loader, java class files misrepresent Kotlin. Kotlin stdlib is available via jetbrains.mps.kotlin.stdlib. That said, direct usage of these classes is only possible from Kotlin code (no BaseLanguage-to-Kotlin interop possible atm). At the same time, Kotlin's Continuation interface is not compatible with Java call mechanism: it's a part of Coroutines framework where it implements the "Continuation-Passing" invocation pattern. In short, Kotlin methods accepting Continuation should not be called from Java directly. 

To be honest, MPS.ThirdParty (and MPS.Kotlin for that matter) is a hack that has no right to exist. IDEA platform is now undergoing a major change and all libraries are being repackaged: instead of being availble in `lib.jar` or other locations, module files are used. We cannot be sure which of these are to be available in the release, this is subject to change, but we keep an eye on that. 

> ` depends on (there is a separate issue that I figured out: #1640) @fisakov You can open the class [CommitCheck](http://127.0.0.1:63320/node?ref=39983771-4e9b-401b-a1a9-1da6c777c843%2Fjava%3Acom.intellij.openapi.vcs.checkin%28MPS.ThirdParty%2F%29%2F~CommitCheck) where the reference to Continuation is broken. All classes from the IntelliJ with references to the standard library can't be used as a result. If this is fixed on MPS master, we can think of merging it but preferable with a migration.

We can look into this issue, but in the meantime keeping MPS.ThirdParty around makes little sense.

fisakov:
I started to uproot MPS.ThirdParty after I found out it was in my way when resolving artifacts from "external layout", which is now a feature supported by build language. Mbeddr core is relying on it (see https://github.com/mbeddr/mbeddr.core/pull/3219). It turned out to be misused: at several locations I found same model imported in parallel both from it and from MPS.IDEA. So, it had to go. 

I updated the changelog with info on MPS.ThirdParty and MPS.Kotlin.

PR #1655, opened 2025-11-11
fix: NullPointerException when opening non-existing node in pagination listener

alexanderpann:


sergej-koscejev (approved):
Ok I'm going to approve this because this is an improvement, whereas the deadlock is impossible to reproduce and the samples are a pain to work with. However, I see two problems remaining in the code (not directly related to the scope of this PR): the use of NavigationSupport and the fact that the editor extensions is registered via the project but editor extensions became application-level some time in or before 2025.1.

PR #1657, opened 2025-11-12
feat(diagram): a flag to suppress autolayout-on-init

sergej-koscejev:
Autolayout on init is run asynchronously and in case of a batch program that attempts to re-layout all diagrams it would cause too many actions queued in EDT that later fail because the editor component was already disposed.

Introduce a flag to disable this behavior in such scenarios.


alexanderpann (approved):
LGTM. The flag doesn't change any existing code.

PR #1683, opened 2025-12-02
perf(linenumbers): they were slowing down the update of the main editor

### LINKED ISSUE #1662: Linenumbers Performance
The line numbers extension significantly slows down the update of the editor. Maybe we can make the computation asynchronous and reduce the delay of the main editor.
I think it's not critical that they are always up-to-date. They can be out of sync with the main editor for a short duration.

slisson:
Expensive computations were done in the paint method which are supposed to be done in the relayout method.

The computation of the lines and their positions is now done only once after 1 second of inactivity. Usually, the inconsistency is only noticeable for the line number at the end. At least the line numbers are not important enough for the user experience to justify more utilization of the EDT thread, which is already an overused resource.

Fixes #1662


slisson:
Did some profiling again with the changes in this PR and there is no relevant time consumed by the linenumbers extensions anymore.

sergej-koscejev:
Please also update the changelog

PR #1694, opened 2025-12-09
feature(intention): Make it possible to customize intention extension

alexanderpann:
All intentions can now be customized through the extension point [IntentionCustomization](http://127.0.0.1:63320/node?ref=r%3A028362d1-b964-410a-a3d5-6096bcd4a2b6%28com.mbeddr.mpsutil.intentions.runtime.plugin%29%2F4784371196443178355) (child filter, description, is applicable block, execute method). Note that you can't customize intentions that are not enabled in the current context (node + editor context) e.g. changing is applicable from false to true. The reason is that we first list the existing available intentions which are already filtered according to the applicable block.


sergej-koscejev:
Looks acceptable but there's a cyclic dependency between the two models in com.mbeddr.mpsutil.intentions.runtime. It's only models, not modules, so it's not a big deal but still let's keep it one-way unless absolutely necessary.

As an idea for future customizations since I see the 'pass map and modify it' pattern taking hold: to take it a step further, instead of passing a map, an even better API IMO would be to pass in a builder object that would have methods to build the map internally or maybe do something else in the future, but that allows to have multiple methods, e.g. as a convenience or when additional features are needed later.

```java
void customizeIntentions(IntentionCustomizationsBuilder builder) {
  builder.customize(node-ptr/MyIntention/, new IntentionCustomization {
    @Override
    public string getDescription(node<> node, EditorContext editorContext, { ⇒ string } originalImpl) {
      "my customized description";
    }
  });

  // later if we e.g. find that users often only customize descriptions and only use hard-coded constants,
  // we could introduce a simpler way without breaking user code:
  builder.modifyDescription(node-ptr/MyIntention/, "my customized description");
}
```

PR #1696, opened 2025-12-09
feat: asynchronous editor updates

slisson:
Usually, you have to be very careful when traversing the model inside editor code because it
introduces dependencies between cells and model elements. A small change in the model can result
in large updates in the editor. For large root nodes the editor can become almost unusable.

The proposed solution is to delay updates of editor cells that are currently not important for the
edited part of the root node and tolerate a short inconsistency.
Most of the time the dependency is either unecessary because MPS just doesn't track them
fine-grained enough, or the change has no effect on the final result, or the affected cells aren't
currently visible because they are outside the viewport.

The cells that are updated immediately are currently determined by which cell is selected.
This heuristic might need to be tweaked after collecting some experience.

There is a new action `Asynchronous Editor Updates` in the context menu of the editor for enabling/disabling this feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project are documented in this file.
 
 The format is *loosely* based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) .The project does *not* follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## December 2025
+
+### Added
+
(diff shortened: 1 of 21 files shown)


slisson:
Moved this feature to the project were it's actually needed. Further development and maintenance will be done there. Sorry for the noise.

sergej-koscejev:
I agree, this is an interesting and potentially valuable feature for other projects as well because editor performance is often a hot topic. But we need to be able to maintain it and update it to newer MPS versions if our customers are to rely on it.

PR #1698, opened 2025-12-10
feat(compare): add 'ignore annotations' property

sergej-koscejev:
* AssertNodeEquals, AreEqualExpression: add 'ignore annotations' property to ignore attributes (annotations) while comparing.
* IDiffable: replace the user object with a transient property. The main reason is actually to get consistent checkboxes in the editor of 'assert node equals'.
* AssertNodeEquals: show bounds of the statement in the editor.
* AreEqualExpression: indicate presence of ignores with an asterisk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ The format is *loosely* based on [Keep a Changelog](https://keepachangelog.com/e
 
 ## December 2025
 
+### Added
+
+- *de.itemis.mps.compare* 'assert node equals' and ':isEqualTo:' expression can now ignore attributes (annotations) when comparing nodes.
+
+### Changed
+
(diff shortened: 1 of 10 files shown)


alexanderpann (changes requested):
- Why is it called attributes and not annotations like in the original language?
- Please add a test that shows that this feature is working.
- The change to the diff checkbox is fine.

sergej-koscejev:
> Why is it called attributes and not annotations like in the original language?

Well, the attributes all extend Attribute, the role they are contained in is called smodelAttribute, so the question should really be why they were called annotations in the current implementation. But I didn't want to break binary compatibility so it was a question of consistency vs correctness here.

> Please add a test that shows that this feature is working.

Will do.

alexanderpann (approved):

Issue #1726, opened 2026-01-30
Building editors with many EnumerationCheckbox instances is too slow


kbirken:
Opening editors where many `EnumerationCheckbox` instances have to be shown may take a lot of CPU time, even if there are only a few different images, which are re-used many times.

Example: The variant configuration editor In IETS3 needs to display one such checkbox for each "line" in the editor (i.e., node of concept `FeatureConfiguration`). The checkboxes all have to show one of the same 8-9 icons, but if there are many of them, creating the editor cell tree takes too long.

kbirken:
PR has been merged successfully to mps20241, closing.

PR #1731, opened 2026-02-02
Improve performance when using many enumeration checkbox widgets in MPS editors

kbirken:
The enumeration checkbox cell will use one AWT Image object for each (see class `EnumCheckboxImage`) enumeration item, i.e. for each potential state of the checkbox. During profiling in a customer project, we measured high CPU usage for building the editor cells for the enumeration checkboxes when many of them (i.e., thousands) are constructed for the same editor pane (for the same root node). This is especially interesting if the enumeration checkboxes are all instances of the same enumeration type with the same icons, i.e., each image is used several thousand times.

As it turns out, the problem is not that the same image is loaded many times. When using `Toolkit.getImage()` ([javadoc](https://docs.oracle.com/javase/8/docs/api/java/awt/Toolkit.html#getImage-java.net.URL-)), the objects returned are using internal caching, so that the same image should only be loaded once if used many times. In a later optimisation, we could think of using `Toolkit.createImage()` and do the caching ourselves to avoid memory leaks. But for now we can let AWT do the caching.

The actual performance problem is retrieving the resource URL for the images many times via the module class loaded of the MPS module runtime (`url = moduleRuntime.getModuleClassLoader().getResource(path)`). Thus, this PR introduces a cache for URLs based on a key pair of module reference and relative image path. In our tests, this cache improved the runtime of the editor buildup by up to 40%.

This solves #1726.

### DIFF
--- a/code/widgets/languages/de.itemis.mps.editor.enumeration/runtime/de.itemis.mps.editor.enumeration.runtime.msd
+++ b/code/widgets/languages/de.itemis.mps.editor.enumeration/runtime/de.itemis.mps.editor.enumeration.runtime.msd
@@ -19,6 +19,7 @@
     <language slang="l:f3061a53-9226-4cc5-a443-f952ceaf5816:jetbrains.mps.baseLanguage" version="12" />
     <language slang="l:fd392034-7849-419d-9071-12563d152375:jetbrains.mps.baseLanguage.closures" version="0" />
     <language slang="l:83888646-71ce-4f1c-9c53-c54016f6ad4f:jetbrains.mps.baseLanguage.collections" version="2" />
+    <language slang="l:f2801650-65d5-424e-bb1b-463a8781b786:jetbrains.mps.baseLanguage.javadoc" version="2" />
     <language slang="l:760a0a8c-eabb-4521-8bfd-65db761a9ba3:jetbrains.mps.baseLanguage.logging" version="0" />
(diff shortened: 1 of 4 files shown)


wsafonov (commented):
As shortly discussed in a huddle, we need to check that the performance issue is not caused by the `ClassLoader.getResource()` call on a ModuleClassLoader that will scan the complete classpath of the module with all the dependencies jars, also including JB runtime libraries. Instead of using this method, invoking `ModuleClassLoader.getOwnResource()` should reduce the search only to a few locations owned by the ModuleClassLoader and thus decreased the lookup time significantly. If the result is roughly comparable with the time measured in case of the custom cache solution, we should rather go this standard way of searching the module resoures.

kbirken:
I checked using `getOwnResource` instead of caching, but it didn't show a significant improvement neither in wall-clock runtime for the diff-usecase nor when analysing it with the (sampling) profiler:

image

So I suggest to add the `getOwnResource(), but still keep the caching in place.

kbirken:
> Before merging, could you please also adjust same things for the bool editor - there we have the same bottleneck if custom images would be used for checkbox states (see ImageLoading class). For now it would be fine just to duplicate the code there.

Done, pls check again.

sergej-koscejev:
@kbirken @wsafonov I don't see any mechanism to get this cache invalidated, beyond restarting MPS. It makes me a bit uneasy. Could we add some kind of listener to invalidate it when a module gets reloaded?

EDIT: or do you think it is unnecessary?

PR #1743, opened 2026-02-16
perf: LayoutInterceptor was traversing the whole cell tree after each editor update

slisson:
In the specific case where I tested this change the editor update delay is reduced from ~200 ms to ~90 ms.


sergej-koscejev (approved):
LGTM. The comment could have stayed.

sergej-koscejev:
I updated it as part of backporting this to 2024.1

PR #1764, opened 2026-03-03
introduce a transaction reconciliation framework for model listeners

dalbrecht1:
## Motivation

Model Listeners sometimes are in a bit of a tricky situation. They can be the most ideal way to define certain kinds of automation, but operate on a quite low level of abstraction and interfere with other routines. In regards to the level of abstraction, I'd compare them to using keymaps in your editors instead of action maps to define user interactions. In regards to the interference, I'd like to point to Modelix, that I had a lot of trouble with at first when applying it to existing projects and workflows recently.

But to be more concrete let me present you with two examples:
1. If a node is moved within the child list of the same parent or even across different parents, then _technically_ this is split into a child-removal and a separate child-addition event when it comes to what model listeners will see. If you now want to or have to perform certain actions on removal to keep your model consistent / error free, then a moved child will erroneously trigger reactions, leaving your model in an altered state after the node-relocation. In most use cases when it comes to child-removal, the followup automation might be to remove additional nodes (like those referencing the now gone node). Now, with a move, these "cleaned up nodes" from the child-removed listener would in the end still be valid and desirable, given that the "referred to" node also still exists, just in a different location. For that, the model listener framework - no matter the power it already has - lacks the options needed to model real developer intent on a higher level of abstraction. Aka, it will not be able to consider the "net result" of a change.
2. Modelix is a powerful and already quite refined way to collaborate on the same project with multiple RCPs (or an RCP and web clients). And while Modelix in its current state even differentiates local edits from remote changes (in that, by triggering an undo you will only revert your own changes, not the ones from colleagues) and can suppress certain kinds of MPS automation in the sync process (e.g. constructors and node factories won't be triggered when the sync creates new nodes on your end), it cannot fundamentally block all kinds of automation from reacting to sync events. Huge thanks to @slisson for his help in understanding and dealing with Modelix intricacies when applying Modelix to our project. That taught me a lot I was able to utilize for this PR.

So, while in both/all cases you should first do your best in optimizing/refactoring model listeners to (a) not be necessary at all, (b) not alter the model in unnecessary ways, and/or (c) be idempotent/re-runnable at least, there are these cases where even the ideal model listener will struggle. For these cases, this PR introduces a utility model, which wants to give more power to the developer in expressing intent in the current model listener framework, by means of simply wrapping your previous listener body in a single, minimalistic "debouncing"-command. Ideally, very little else has to change in the actual code.

## Proposal

Most if not all user interactions are encapsuled in a command or an undo-transparent action (I'll use "command" as the shorthand going forward, but remember that this means these two types of transactions). The same is true for any operation Modelix takes while fetching and integrating external changes into your local project. Hence, the MPS platform already "knows", that there are more atomic changes to come before the whole transaction finalizes. Therefore, within the model listener you are dealing with an intermediate state, but the ultimate state after the whole transaction can look significantly different. Hence, model listeners only see a snapshot at the exact times certain operations are performed, but do not consider a change in the big picture of the whole command. And while you certainly could just make use of "invokeLater"-calls or spin up your own boilerplate solution to apply in all model listeners, the strength of this proposal is a complete "action reconciliation runtime" that comes at almost no cost to you as the developer or the runtime (with no proof for that claim so far).

Whenever a model listener hook is executed, and the hook is executed in the context of a transaction that can have "user intent", you are now able to not run your code right away, but have it scheduled to only run at the end of the current transaction, when this delayed execution (the debouncing principle) will allow your code to consider the entirety of the operations being included in the command.

## Usage Example

Imagine the context of a sort of truth table like use case. But one where there isn't just "true" and "false", but where the user can specify any terminals they like. But based on these terminals the user also should be able to model a binary operator in the form `terminal op terminal => terminal`. Here, if the table editor is done right, you do not have to take any actions when a new terminal is added. However, you might have to remove any "truth table entries" that feature a specific terminal, if the user wants to delete this particular terminal.

Hence, you might have a model listener like:
```java
child removed in role terminals (instance, child)->void {
	findTruthTable().tableEntries.removeWhere({~it => it.lhs :eq: child || it.rhs :eq: child || it.result :eq: child; });
}
```

But if a terminal is just manually reordered, then the table will still drop these entries regardless. With this new framework, you can instead do the following:
```java
child added in role terminals (instance, child)->void {
	TransactionReconciler.getInstance(instance).cancelByOrigin(child, "truthTable");
}
child removed in role terminals (instance, child)->void {
	final SNodeId childId = child/.getNodeId();
	TransactionReconciler.getInstance(instance).schedule(child, "truthTable", {node<> node =>
		findTruthTable().tableEntries.removeWhere({~it => 
			it/.getReference(link/TruthTableEntry : lhs).getTargetNodeId() :eq: childId ||
			it/.getReference(link/TruthTableEntry : rhs).getTargetNodeId() :eq: childId ||
			it/.getReference(link/TruthTableEntry : result).getTargetNodeId() :eq: childId; });
	});
}
```

As you see, you receive one benefit at the cost of putting yourself in the timeline where the node is already gone for good, hence having to shift to checking the underlying node-ids. Which you could circumvent by pre-querying the entries-to-be-removed prior to scheduling the removal and within that callback to simply detach them all. I'll show an example of that in the following code snippet.

Or, using another feature of this framework, you can make use of a "strategy" to already express intent of "only run on actual removal", like that:
```java
// note, that you no longer need a 'child added' listener here to cancel
child removed in role terminals (instance, child)->void {
	final list<node<ThruthTableEntry>> toBeRemoved = findTruthTable().tableEntries.where({~it => it.lhs :eq: child || it.rhs :eq: child || it.result :eq: child; }).toList();
	TransactionReconciler.getInstance(instance).schedule(child, "truthTable", IReconciliationStrategy.IF_ABSENT, {node<> node =>
		toBeRemoved.forEach({~it => it.detach; });
	});
}
```

And the final gap between the original listener and the last one could even be bridged by defining specific callback types that capture even more intent (the one used here is already included in the framework, but they are just implementations of a given (functional) interface). Then it really looks very much the same as before.
```java
child removed in role terminals (instance, child)->void {
	TransactionReconciler.getInstance(instance).schedule(child, "truthTable", IReconciliationStrategy.IF_ABSENT, new ReconciliationAction.DetachNodesAction(
		findTruthTable().tableEntries.where({~it => it.lhs :eq: child || it.rhs :eq: child || it.result :eq: child; }).toArray
	));
}
```

If at any point this new delaying engine isn't able to retrieve a fitting instance or finds itself outside a command - aka in a situation where there is no broader "user intent" to consider - the schedule method will simply immediately execute the provided hook, should the strategy call for it. And if you want to perform specific actions on move- and remove-only, you are free to include both definitions in the same model listener. Then, either on command finalization, or right away, only the "correct" one is executed. This ensures that the utility is safe to use in unit tests or headless scripts as well where a Command context might not be present, maintaining the 'no callback lost' guarantee.

## Further Work

This PR for now only proposes a "code/runtime only" solution which expects the developer to call the right classes and methods to let their callbacks be managed by my new system. Functionality-wise this might already be as refined as it has to be, though I'm open for any comments and discussions. But what this proposal still lacks in my eyes is a nice syntax that highlights the intent better. See, I spent some time optimizing this such that the intent of the written listener code is better captured at runtime, but at the cost of "silly static calls" that are meaningless to the uninitiated.

And while it might not be much work to elevate the principles to an intent-focused syntax/DSL, I just lacked the time right now and first wanted to let you give feedback on the principle behind my concept. But in the end, I'd wish for a syntax that is closer to the "command / read action / write action with repository" statements or the "with _ models in scope do" statement. Which should be a somewhat trivial translation task, I assume.

One example, where such a DSL could still improve on the current state, is to create meaningful aliases for the strategies that adapt to the type of listener. While technically the "IF_ABSENT" strategy perfectly captures the state of the node, within a "child-removed" listener and to better communicate intent, the same states could be made available as "node moved" vs "node removed" strategies within the DSL.

### DIFF
--- a/code/.mps/modules.xml
+++ b/code/.mps/modules.xml
@@ -99,6 +99,7 @@
       <modulePath path="$PROJECT_DIR$/modellisteners/com.mbeddr.mpsutil.modellisteners.runtime/com.mbeddr.mpsutil.modellisteners.runtime.msd" folder="modellisteners" />
       <modulePath path="$PROJECT_DIR$/modellisteners/com.mbeddr.mpsutil.modellisteners.sandbox/com.mbeddr.mpsutil.modellisteners.sandbox.msd" folder="modellisteners" />
       <modulePath path="$PROJECT_DIR$/modellisteners/com.mbeddr.mpsutil.modellisteners.sandboxlang/com.mbeddr.mpsutil.modellisteners.sandboxlang.mpl" folder="modellisteners" />
+      <modulePath path="$PROJECT_DIR$/modellisteners/com.mbeddr.mpsutil.modellisteners.utility/com.mbeddr.mpsutil.modellisteners.utility.msd" folder="modellisteners" />
       <modulePath path="$PROJECT_DIR$/modellisteners/com.mbeddr.mpsutil.modellisteners/com.mbeddr.mpsutil.modellisteners.mpl" folder="modellisteners" />
       <modulePath path="$PROJECT_DIR$/modelmerger/languages/de.itemis.mps.modelmerger/de.itemis.mps.modelmerger.mpl" folder="modelmerger" />
       <modulePath path="$PROJECT_DIR$/modelmerger/languages/de.itemis.mps.modelmerger/sandbox/de.itemis.mps.modelmerger.testhelper.msd" folder="modelmerger" />
--- a/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
+++ b/code/build/solutions/de.itemis.mps.extensions.build/models/de.itemis.mps.extensions.build.mps
@@ -14641,6 +14641,71 @@
           </node>
         </node>
       </node>
+      <node concept="1E1JtA" id="4kz7QXL0NP1" role="2G$12L">
+        <property role="BnDLt" value="true" />
+        <property role="TrG5h" value="com.mbeddr.mpsutil.modellisteners.utility" />
+        <property role="3LESm3" value="cd9c13ae-91f8-404c-abd5-3b0b798fc9f6" />
+        <node concept="398BVA" id="4kz7QXL0NWW" role="3LF7KH">
+          <ref role="398BVh" node="2fo8bJE$D4t" resolve="extensions.code" />
+          <node concept="2Ry0Ak" id="4kz7QXL0Ok_" role="iGT6I">
+            <property role="2Ry0Am" value="modellisteners" />
+            <node concept="2Ry0Ak" id="4kz7QXL0O$m" role="2Ry0An">
+              <property role="2Ry0Am" value="com.mbeddr.mpsutil.modellisteners.utility" />
+              <node concept="2Ry0Ak" id="4kz7QXL0OO7" role="2Ry0An">
+                <property role="2Ry0Am" value="com.mbeddr.mpsutil.modellisteners.utility.msd" />
+              </node>
+            </node>
+          </node>
+        </node>
+        <node concept="1SiIV0" id="4kz7QXL0PsS" role="3bR37C">
+          <node concept="3bR9La" id="4kz7QXL0PsT" role="1SiIV1">
+            <ref role="3bR37D" to="ffeo:1H905DlDUSw" resolve="MPS.OpenAPI" />
+          </node>
+        </node>
+        <node concept="1SiIV0" id="4kz7QXL0PsU" role="3bR37C">
+          <node concept="3bR9La" id="4kz7QXL0PsV" role="1SiIV1">
+            <ref role="3bR37D" to="ffeo:mXGwHwhVPj" resolve="JDK" />
+          </node>
+        </node>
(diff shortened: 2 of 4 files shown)


sergej-koscejev:
Without looking at the code yet, MPS provides a batch event listener ([RepositoryChangeTracker](https://github.com/JetBrains/MPS/blob/master/core/smodel/source/jetbrains/mps/smodel/event/RepositoryChangeTracker.java)) which you can use to react to all events that happened in a command. Are you aware of it? Would it help in your case or do you need to react to events during a command? See [UsageModelTracker](https://github.com/JetBrains/MPS/blob/master/workbench/mps-refactoring/jetbrains.mps.ide.refactoring.platform/source_gen/jetbrains/mps/ide/platform/refactoring/UsagesModelTracker.java) for a (trivial and only) usage example.

sergej-koscejev:
Closing this for now due to no activity, please reopen or recreate when you have an example or tests.

PR #1784, opened 2026-03-23
Support for adding lifecycle methods to editor tests

sergej-koscejev:
Add an attribute to provide lifecycle (before/after tests) methods to editor test cases.

Also rewrite the generator of FileNodeEditorExpression without reflection. The type of the expression is changed from MPSFileNodeEditor to the more abstract FileEditor.


wsafonov (commented):
The new lifecycle methods for the editor tests looks good. I have only one remark regaring `FileNodeEditorExpression`. I'm not sure if any client code specifically required `MPSFileNodeEditor` instead of simple `FileEditor`. But since we can probably safely assume that in the tests context we would deal anyway only with `MPSFileNodeEditor`, maybe for convenience we should check the returned file editor and automatically cast it to `MPSFileNodeEditor` in the generated code. Then there will be for sure no adjustments needed in any client code. WDYT?

sergej-koscejev:
I asked @alexanderpann, he said he only added it for an example undo/redo editor test where FileEditor is enough.

sergej-koscejev:
But I think I'll just add the cast there, to be careful... I already added it, then removed for simplicity, but since it raises eyebrows I'll add it back.

wsafonov (approved):
Alright, thanks for the adjustment. Looks good now.

PR #1786, opened 2026-03-25
fix(celllayout): LayoutInterceptor wasn't installed in some cases

slisson:
Fixes a regression introduced by https://github.com/JetBrains/MPS-extensions/pull/1743


wsafonov:
Will need to test this first in the customer project under 2024.1 to confirm that the regression has been resolved completely.

wsafonov:
Based on tests in the customer project, the grid layout seems to work there as before. However, when I checked the behavior in the `TestCase_GridLayoutFlatten` test editor, the results look pretty different before the #1743 and after this patch: 


| **Before:** | **After:** |
| :---: | :---: |
| image | image |



The new layouting behavior looks pretty broken for that specific test grid. There seems to be still another part of the interceptor logic missing an install on some cells. Could you please double-check this as well?

slisson:
That's also fixed now.

wsafonov (approved):
Fixed indeed. Also great that you have added now some tests to avoid such unpleasant regressions in the future. Thanks!

PR #1801, opened 2026-04-27
linenumbers: silence 'Do not create alarm without coroutineScope' warning

sergej-koscejev:
We still create alarm without coroutineScope because we don't want to mess with Kotlin. Passing ThreadToUse explicitly at least disables the warning.


wsafonov (approved):
`ThreadToUse.SWING_THREAD` seems legit as the default for the Alarm, which expects `updateNow()` to be executed on EDT.  Approved.

PR #1824, opened 2026-06-15
fix(shadowmodels): only run commands against an initialized project

sergej-koscejev:
Running commands against an uninitialized project causes an unrelated third-party plugin to fail because it attaches a command listener that then fires too early and tries to retrieve a component from the uninitialized project, causing an exception.


wsafonov (approved):
The change seems reasonable. Approved.

PR #1827, opened 2026-06-17
fix(tables): AssertionException in EditorCell_GridCell.paint for head…

slisson:
…less components

The sticky headers feature tries to read the scroll position which isn't available in case of headless components. The screenshots can then not be created by the generator.


arimer (approved):
Change look good, thanks for updating the changelog.

PR #1833, opened 2026-06-30
Fix navigation deadlocks

sergej-koscejev:


wsafonov (approved):
Looks good. Additionally, I fixed the layout of the search buttons to make them aligned with the page buttons.

PR #1838, opened 2026-07-15
perf(tables): support incremental update

slisson:
Removes the EditorCacheHacks.noCaching workaround. It's stays enabled for tables containing annotations because thats more complicated to fix but a rarely used feature.

In addition to the newly added tests that check the incremental update behavior I also manually checked the existing table implementations in mbeddr and IETS3.

This change should significantly improve the responsiveness of large tables.


dbinkele:
@slisson please add a changelog entry.

dbinkele (approved):
LGTM

PR #1845, opened 2026-07-28
BaseLanguageExtensions: improve group by operation

alamqadem:


wsafonov:
The logic looks solid - thanks for the contribution!

I do have one thought regarding the UX. Since `groupBy` now accepts two additional optional functional parameters, it becomes less obvious at a glance which closure corresponds to which parameter. Would it make sense to use more descriptive closure parameter names instead of the generic `it`, for example `value` for the value selector function and `group` for the aggregate function?

It could also be helpful to implement `parameters-information` so users can quickly see the current signature in a parameter tooltip - with the selected closure parameter highlighted using the `Ctrl+P` shortcut.

PR #1849, opened 2026-07-31
perf(grammarcells): parse with a chart parser instead of backtracking

slisson:
## What

Replaces the parsing algorithm behind grammar cells with a chart parser (Earley). `Parser` keeps its API; the four new classes in `com.mbeddr.mpsutil.grammarcells.runtime` are `ChartParser` (the engine) and `ChartItem` / `ChartLink` / `ChartState` (the chart).

## Why

`Parser` tried every way of distributing the input tokens over the symbols of a rule (`InputCursor.iterateRanges`) and re-parsed the same token ranges once per distribution, with nothing memoised — the work grew exponentially with the number of tokens. The step budget (`DEFAULT_TIMEOUT`), the `ParserFailureNotifier` extension point and the recursion depth cutoff of 20 were all there to contain that.

A chart parser enters every rule that can start at an input position into the chart once, and reuses a subtree parsed for one rule in every other rule that asks for the same tokens, which makes the work polynomial. The links between chart items form the parse forest, so an ambiguous input keeps all its readings and the tree is chosen afterwards.

## Behaviour that is deliberately kept

- the rule prefilter (a rule whose constants do not all appear in the input cannot match) and its ordering: rules producing a concept the input was read from first, then by priority;
- the priority and associativity check — the same `canBeRotated` logic, now asked where a child is attached to its parent instead of before descending, and cached per (parent rule, symbol index, child rule);
- a child symbol matched by a single untouched `ChildToken` still wins over re-parsing those tokens;
- the preference for the concept a token was read from, now evaluated per span.

Two things change on purpose: the depth cutoff of 20 is gone (a chart does not recurse), and the step budget now counts chart items rather than descent steps. The recursive descent stays in `Parser` because `ChildSymbol.parse` still offers it as the way to parse a range.

## Numbers

Measured on the grammarcells sandbox grammar, unparseable input (blocks of `[ ( n + 2 * 3 ) ? 4 : 5 ]` joined by `+`, trailing `+`):

| tokens | before | after |
| --- | --- | --- |
| 42 | 23 ms | 0 ms |
| 56 | 69 ms | 0 ms |
| 70 | 285 ms | 0 ms |
| 84 | 692 ms | 1 ms |
| 168 | aborted after 20M steps | 0 ms |

A valid 321-token expression parses in 53 ms with 62k chart items; chart size grows polynomially (28 tokens → 466 items, 56 → 931, 112 → 1921, 224 → 4141).

## How it was checked

The grammar cells test suites were run in MPS after the change — **all green**.

Beyond that, the old and the new implementation were compared on identical input, all producing identical parse trees:

- 483 generated inputs (valid expressions plus one-token deletions and insertions, up to ~35 tokens): 483/483 identical, 231 ms → 16 ms in total;
- a 342-case corpus recorded before the change and re-run after: output byte-identical;
- the 76 token sequences read out of the sandbox models with `Parser.getTokens` (so `ChildToken` / `PropertyToken` and the original-concept preference are exercised): identical.

Eleven sandbox nodes reparse to a different shape than the one stored in the model, but both implementations agree on all of them — those models hold right-leaning trees that either parser rebalances.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/code/grammarcells/solutions/com.mbeddr.mpsutil.grammarcells.runtime/models/com/mbeddr/mpsutil/grammarcells/runtime.mps
+++ b/code/grammarcells/solutions/com.mbeddr.mpsutil.grammarcells.runtime/models/com/mbeddr/mpsutil/grammarcells/runtime.mps
@@ -328,6 +328,7 @@
         <child id="8276990574909234106" name="finallyBody" index="1wplMD" />
       </concept>
       <concept id="1073239437375" name="jetbrains.mps.baseLanguage.structure.NotEqualsExpression" flags="nn" index="3y3z36" />
+      <concept id="1081855346303" name="jetbrains.mps.baseLanguage.structure.BreakStatement" flags="nn" index="3zACq4" />
       <concept id="1184950988562" name="jetbrains.mps.baseLanguage.structure.ArrayCreator" flags="nn" index="3$_iS1">
         <child id="1184951007469" name="componentType" index="3$_nBY" />
         <child id="1184952969026" name="dimensionExpression" index="3$GQph" />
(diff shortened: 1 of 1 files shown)


slisson:
Was just a test.

PR #1865, opened 2026-08-18
fix(richtext): new Word instances are initialzed with an empty string…

slisson:
… again instead of a space

9dc8b0219b9ffa919986f16ce333888d0a879717 changed the constructor of Word to initialize the text with a single space. The commit messages is about the a different change and there is no reasoning anywhere for this change. https://github.com/JetBrains/MPS-extensions/pull/805 has only a title "Richtext improvements" and no description.

If this is supposed to be an improvement then it's too broad and should be done somewhere else. 00447ba0342d0ac5f216f49495375d0c81ae157a even had to work around this unjustified change of behavior. In a customer project tests are failing because every new Text is now initialized with a single space instead of being empty.

Reverted this change since there is no reason given for it and there are no tests that demonstrate the intended "improvement".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project are documented in this file.
 
 The format is *loosely* based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) .The project does *not* follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## August 2026
+
+### Fixed
+
+- *de.slisson.mps.richtext* Word instances were initialized with a single space. Restored previous behavior of initializing them with an empty string.
+
 ## July 2026
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


sergej-koscejev:
@alexanderpann do you remember the reason for this change?

alexanderpann:
I think it was an attempt to fix a usability issue where a new word is created, and no text could be entered because there was no string yet. My change was a big mistake and should be reverted, or an empty string should be used. It caused some issues with jumping cursors, and the whitespaces were inserted at surprising/unnecessary positions.
FYI: In the past, a few tests in an upstream platform failed as a result of the change, so I assume that you will run into the same failures again when you change the behavior.
Ideally, this fix should be done in an older maintenance branch because it affects all branches from at least MPS 2022.2.

sergej-koscejev (approved):

sergej-koscejev:
Approving this for 2025.1 and leaving the decision on whether to fix in earlier versions to Sascha.

arimer:
Alex P. forecast was right: we have failing (expected) tests in iets3: https://github.com/IETS3/iets3.opensource/pull/1949/changes

PR #1886, opened 2026-09-03
docs: fix typos and copy-paste errors in the extension documentation

Avicennasis:
## Summary

Fix 20 typos, grammar slips and copy-paste errors in the extension documentation under `docs/`. Documentation only - no code or MPS models touched.

**docs/extensions/editor/tables.md**
- `- **disable right row end cells**: there's a special cell to the left of table rows` -> `to the right of table rows` (copy-paste from the "disable left row end cells" bullet directly above)
- `a function that initialized the shared variables` -> `initializes`
- `an even lower level form of declaration a table` -> `declaring a table`
- `setting the cells directely in the grid` -> `directly`

**docs/extensions/editor/math.md**
- `First you have to add some child cells. The need to have a name` -> `They need to have a name`
- `After settings some variables` -> `After setting`
- ``The `paint`method`` -> ``The `paint` method``
- `math.sum` and `math.product` had their link labels swapped: `math.product` now points at **pi notation** and `math.sum` at **sigma notation**. `math.sum` also had its URL corrected to `.../sigma-notation/` (verified 200) since it previously pointed at the pi-notation article.

**docs/extensions/editor/grammar-cells.md**
- `The editor cell of th child` -> `of the child`

**docs/extensions/editor/celllayout.md**
- `grow the cell to the with of the parent` -> `to the width of the parent`

**docs/extensions/editor/richtext.md**
- `there are can't be two consecutive Words` -> `there can't be`

**docs/extensions/utils/model-merger.md**
- `concepts wihtout policies otherwise an exception will be throw` -> `without policies ... will be thrown`
- `as described bellow` -> `as described below`
- `- **Right**: use the value of the right property, and discard the value of the right property.` -> `discard the value of the left property` (the Children and Reference sections below already use the correct left/right pattern)
- `and mde\`de.itemis.model.merge\` language` -> ``and the `de.itemis.model.merge` language``
- `defined in the configuraiton` -> `configuration`
- `Additonaly one can use` -> `Additionally`

**docs/extensions/utils/node-versioning.md**
- `the users wants to detect changes` -> `the users want`; `Now when users wants check` -> `Now when users want to check`
- `might under go substantial changes` -> `undergo`

### Deliberately not changed (verified, not typos)

- `README.md`: `Mbeddr2_Mbeddr_Gradle_MpsExtenstions` looks like a misspelling but it is the **real TeamCity build configuration id**. Requesting the status icon for that id returns a live build status, while the "corrected" spelling returns *no permission to get data*. Fixing it would break the badge.
- `math.md`: `side-tranformation-helper-cells` is the literal cell name in `code/math/languages/de.itemis.mps.editor.math/languageModels/editor.mps`, so the docs are correctly quoting the identifier.
- `node-versioning.md`: `` `nodecomparator` `` is left as-is - it reads as a reference to the comparator extension rather than prose.

### One thing you may want to look at

`docs/extensions/editor/celllayout.md` ends mid-sentence: the file's last line is *"...it is very likely that"* with nothing after it. It was left untouched because completing it would mean inventing content, but it looks like text was lost.

### DIFF
--- a/docs/extensions/editor/celllayout.md
+++ b/docs/extensions/editor/celllayout.md
@@ -60,7 +60,7 @@ installed recursively when any style property of the language is used.
 
 ### Grow and Push
 
-- **grow-x** is a flag that tells the layouter to grow the cell to the with of the parent.
+- **grow-x** is a flag that tells the layouter to grow the cell to the width of the parent.
 - **push-x** is a property that has the same effect as setting `grow-x` on the cell and all ancestor cells.
 - **grow-y** is a flag that tells the layouter to grow the cell to the height of the parent.
 - **push-y** is a property that has the same effect as setting `grow-y` on the cell and all ancestor cells.
--- a/docs/extensions/editor/grammar-cells.md
+++ b/docs/extensions/editor/grammar-cells.md
@@ -52,7 +52,7 @@ A cell that supports entering an optional text to add an optional containment li
 ## grammar.wrap
 
 A cell that lets you enter a child in a context where a parent is expected (e.g. enter the type of a 
- variable declaration to create the declaration itself). The editor cell of th child is wrapped and the instantiation of the
+ variable declaration to create the declaration itself). The editor cell of the child is wrapped and the instantiation of the
  child is triggered. It can be used for containment links of children with mandatory (1) or optional (0..1) cardinality.
 
 ### Supported properties (side transformations)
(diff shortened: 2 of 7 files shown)


sergej-koscejev (approved):

sergej-koscejev:
Thanks!

